PR #11, opened 2024-09-07
misc tool configuration

harrysolovay:
Configured dprint and cspell. Did not yet configuration a checks workflow.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,7 +9,7 @@
     "editor.defaultFormatter": "dprint.dprint"
   },
   "[rust]": {
-    "editor.defaultFormatter": "dprint.dprint"
+    "editor.defaultFormatter": "rust-lang.rust-analyzer"
   },
   "[toml]": {
     "editor.defaultFormatter": "dprint.dprint"
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,22 @@
+{
+  "[json]": {
+    "editor.defaultFormatter": "dprint.dprint"
+  },
+  "[jsonc]": {
+    "editor.defaultFormatter": "dprint.dprint"
+  },
+  "[markdown]": {
+    "editor.defaultFormatter": "dprint.dprint"
+  },
+  "[rust]": {
+    "editor.defaultFormatter": "dprint.dprint"
+  },
+  "[toml]": {
+    "editor.defaultFormatter": "dprint.dprint"
+  },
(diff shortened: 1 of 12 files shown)


dprint.json
```diff
@@ -0,0 +1,23 @@
+{
+  "incremental": true,
+  "indentWidth": 2,
+  "lineWidth": 120,
+  "exec": {
+    "commands": [
+      {
+        "command": "rustfmt --config group_imports=StdExternalCrate,imports_granularity=crate",
```
harrysolovay, line 8: Some `group_imports`-sorting is better than none imo. Thoughts?

tjjfvi: The issue I have with it is that it categorizes other members in the workspace as external. I'm good with `imports_granularity=crate`, though.

You mentioned something about needing to use the cli to pass those options, but afaict that is not the case; they can just be specified in `rustfmt.toml` (and that would be my preference).

harrysolovay: Rustfmt didn't apply the `imports_granularity` option when specified via the `rustfmt.toml` (for me at least). I'd imagine this is why it's been specified via `exec` [here in `dprint`](https://github.com/dprint/dprint/blob/29b203c2bc09cb05ced99e4de46b7b0a55fa427e/dprint.json#L6) as well.

tjjfvi (approved):

dprint.json
tjjfvi: Perhaps your editor is using an outdated version of `rustfmt`

.vscode/settings.json
```diff
@@ -0,0 +1,22 @@
...
+    "editor.defaultFormatter": "dprint.dprint"
+  },
+  "[jsonc]": {
+    "editor.defaultFormatter": "dprint.dprint"
+  },
+  "[markdown]": {
+    "editor.defaultFormatter": "dprint.dprint"
+  },
+  "[rust]": {
+    "editor.defaultFormatter": "dprint.dprint"
```
harrysolovay, line 12:
```suggestion
    "editor.defaultFormatter": "rust-lang.rust-analyzer"
```

dprint.json
harrysolovay: Ah, my toolchain was on stable, not nightly. Setting to nightly fixed it.

tjjfvi: Yeah, though I try to support nightly for usage of the program, some of the development relies on nightly features and that's fine IMO.

PR #31, opened 2024-09-29
vine: the inverse operator

tjjfvi:
The inverse operator, `~`, is a fundamental primitive for Vine; features like mutable references and functions can both be implemented with it.

This also introduces the concept of a *space*, which is the inverse of a value; it is a space where a value can be put.

A place consists of a value and a space.

This concept of a space fills in some existing gaps in the language. For example, the assignment statement is now of the form `space_expr = value_expr`. (Previously, it was handled as `value_pat = value_expr`, like let statements, which is very incorrect.)

Similar to how the reference operator takes a place and wraps it into a value, the inverse operator can be thought of taking a space and wrapping it into a value. The inverse operator can also work on values, unwrapping them into spaces. When passed a place, the inverse operator returns another place, but the space is wrapped to be the value of the new place, and the value is unwrapped to be the space of the new place.

Also important to note is that if the inverse operator is passed a space of type `T`, it outputs a value of type `~T`. (At least, conceptually; of course Vine does not yet have a formal type system.)

The inverse operator is an involution; `~~x` is always equivalent to `x`.

At the Ivy level, the inverse operator is a no-op; it simply changes the way things are wired. It does, however, mean that Ivy nets are now non-polarizable. (This could be rectified, if needed, by representing the inverse with a unary node.)

See the `inverse.vi` test for examples of using the inverse operator to implement mutable references, functions, and variables with reversed causality.


harrysolovay (approved):

vjjft (approved):
*reviewed with assumed ability*

PR #41, opened 2024-10-15
vine: boolean expressions

tjjfvi:
Implements:
- `&&` and `||` (with proper short-circuiting behavior)
- `!` (was previously bitwise not; now logicial)
- `is` expressions
  - these replace/generalize `if let`/`while let` (and chains thereof)
    - see also https://github.com/rust-lang/rfcs/pull/3573


vjjft (approved):
looks good (especially with that commented test)

Issue #7, opened 2024-09-06
std: various APIs


tjjfvi:
- [X] `Result` #42 
- [x] `Array` #113 
  - binary tree for `O(log n)` access
- [x] `List` #46 
  - linked queue for `O(1)` concat
  - literal syntax exists; need method support
- ...?

harrysolovay:
For inspiration: https://github.com/denoland/std

PR #42, opened 2024-10-18
std: result

stevenhuyn:
Addresses part of #7 

For `flatten` I followed: https://doc.rust-lang.org/std/result/enum.Result.html#method.flatten kinda, what about the error case cause it's not strongly typed?

nws if you're not taking any PRs


tjjfvi:
Looking good; thanks for the PR! I'm going to tweak the example some to also show implementing a primality test in vine – I like the idea of using that test to showcase `Result`.

tjjfvi (approved):
Thanks!

PR #48, opened 2024-10-21
vine: add continue expression

stevenhuyn:


vine/src/compile.rs
```diff
@@ -54,6 +54,7 @@ pub struct Compiler<'n> {
 
   return_target: Option<(Local, ForkId)>,
   break_target: Option<ForkId>,
+  loop_target: Option<(ForkId, StageId)>,
```
tjjfvi, line 57: since the `ForkId` of `break_target` will always be the same as of `loop_target`, so `break_target` can be removed

tjjfvi (approved):
looks great; thanks!

PR #64, opened 2024-10-22
test: Remove orphaned snapshots

### LINKED ISSUE #49: snapshots should detect when there snapshots with no corresponding test
Renaming a test will leave old snapshot files committed, `test.rs` should detect this.

Potentially we need a `cargo test` vs a `cargo test -- -update` as well.

stevenhuyn:
This PR will now make any invocation of `cargo test` remove any snapshots with no corresponding test file.

We have moved the `vine/fail` tests to its own folder `vine_fail`

Closes #49


stevenhuyn:
Proof of it working on these 2 commits:
image

tests/tests.rs
```diff
@@ -106,6 +137,50 @@ fn run_iv(group: &str, name: &str, path: &str, input: &[u8], output_ext: &str) {
...
+fn remove_orphan_snapshots(path: &'static str, names: &[String]) {
+  let Ok(entries) = fs::read_dir(path) else {
+    println!("Cannot read path: {}", path);
+    return;
+  };
+
+  let mut orphans = false;
+  for entry in entries {
+    let Ok(entry) = entry else {
+      continue;
```
tjjfvi, line 151: if you remove the semicolon here (and on the other `continue`s), rustfmt should collapse these to a single line, which I think looks a bit nicer

tests/tests.rs
```diff
@@ -106,6 +137,50 @@ fn run_iv(group: &str, name: &str, path: &str, input: &[u8], output_ext: &str) {
...
+    };
+
+    let Some(file_name_str) = file_name.to_str() else {
+      continue;
+    };
+
+    let file_name = file_name_str.to_owned();
+    if !names.contains(&file_name) {
+      orphans = true;
```
tjjfvi, line 165: When there are orphaned snapshots, instead of always deleting the files and always failing, can you branch on `should_write_snapshot()` and either delete the snapshots and succeed, or fail and leave the snapshots be?

tests/tests.rs
```diff
@@ -20,51 +20,74 @@ fn tests(t: &mut DynTester) {
...
+
+  t.group("remove_orphan_snapshots", |t| {
+    t.test("ivy", move || {
+      remove_orphan_snapshots("tests/snaps/ivy", &ivy_tests);
+    });
+    t.test("vine", move || {
+      remove_orphan_snapshots("tests/snaps/vine", &vine_tests);
+    });
+    t.test("vine_fail", move || {
+      remove_orphan_snapshots("tests/snaps/vine_fail", &vine_fail_tests);
```
tjjfvi, line 73: can you change this to `fail/vine`? should make tab completing nicer

tests/tests.rs
stevenhuyn: very neat

stevenhuyn:
It's not very robust atm, so I'm thinking if there's a nice way to do it without pulling in a dependency like https://docs.rs/ignore/latest/ignore/

tjjfvi:
IMO it's fine for it to depend on the non-existence of other things in those directories

PR #72, opened 2024-10-23
vine: add `break` with value

### LINKED ISSUE #66: vine: `break` with value
In `loop`s, `break` could be supplied a value, which would become the result of the `loop` expression.

The most complicated part of this would likely be the parsing, as there's not currently a mechanism to optionally parse an expression. (Though this would be useful for `return` as well.)

stevenhuyn:
Closes #66


stevenhuyn (commented):
fib_repl example is broken atm

vine/examples/break_return.vi
```diff
@@ -0,0 +1,20 @@
...
+
+fn unit() {
+  return;
+}
+
+fn loop_unit() {
+  return loop {
+    break "hey";
+  }
+}
```
stevenhuyn, line 20: newline

tests/snaps/vine/fib_repl/output.txt
```diff
@@ -1,16 +1 @@
...
-> 13
-> 21
-> 34
-> 55
-> 10946
-> 3314859971
-> 2458604770
-> invalid number
-> invalid number
 >
```
stevenhuyn, line 1: what is causing this

vine/src/parser.rs
```diff
@@ -248,10 +248,24 @@ impl<'ctx, 'src> VineParser<'ctx, 'src> {
...
+        })));
+      }
       return Ok(ExprKind::Return(Box::new(self.parse_expr_bp(BP::ControlFlow)?)));
     }
     if self.eat(Token::Break)? {
-      return Ok(ExprKind::Break);
+      if self.check(Token::Semi) {
+        self.eat(Token::Semi)?;
+        return Ok(ExprKind::Return(Box::new(Expr {
+          span: self.span(),
```
stevenhuyn, line 264: 2 things feel off here:
- the span
- check() then eat()

vine/src/compile/build_stages/control_flow.rs
```diff
@@ -95,9 +95,11 @@ impl Compiler<'_> {
     Port::Erase
   }
 
-  pub(super) fn lower_break(&mut self) -> Port {
-    let (fork, _stage) = self.loop_target.unwrap();
-    self.diverge(fork);
+  pub(super) fn lower_break(&mut self, r: &Expr) -> Port {
+    let r = self.lower_expr_value(r);
+    let ret = self.return_target.unwrap();
```
tjjfvi, line 100: this is causing the `fib_repl` failure; `break` is now identical to `return`

vine/src/parser.rs
tjjfvi: `eat` returns a boolean, so it could just be `if self.eat(Token::Semi)? { ... }`. though for this specific case, I don't think it should eat the `;`, as this would make something like `break; + 1` a valid expression; it should leave the `;` there to be processed by `parse_stmt`

tjjfvi: And yeah, the span is strange here. I think it might be cleaner here for it to be an `Option<Expr>`, rather than using `()` like I suggested.

vine/src/compile/build_stages/control_flow.rs
stevenhuyn: yes break_target now needs a local

stevenhuyn: but why would fib_repl fail if lower_break never gets called (my assumption given there's no break command in it)

tjjfvi: `fib_repl` calls `std::io::read_line`, which uses `break`

stevenhuyn: (i swear I added a println to see)

tjjfvi: (by default the compiled code gets sent to stdout so you might have more luck with `eprintln!`)

tests/snaps/vine/fib_repl/output.txt
stevenhuyn: bad return bruh

tjjfvi (approved):
Looks good; thanks!

PR #73, opened 2024-10-24
vine: allow returns with no expression

stevenhuyn:
Should be merged before #72 
Addresses a nice to have from #66


vine/src/compile/build_stages.rs
```diff
@@ -132,7 +132,11 @@ impl Compiler<'_> {
       ExprKind::Loop(body) => self.lower_loop(body),
       ExprKind::While(cond, body) => self.lower_while(cond, body),
       ExprKind::If(cond, then, els) => self.lower_if(cond, then, els),
-      ExprKind::Return(r) => self.lower_return(r),
+      ExprKind::Return(None) => {
+        let unit = Expr { span: Span::NONE, kind: ExprKind::Tuple(vec![]) };
+        self.lower_return(&unit)
```
tjjfvi, line 137: I think it would be cleaner for `lower_return` to take an `Option<Expr>` and just not write to the local if it's `None`

vine/src/parser.rs
```diff
@@ -248,7 +248,11 @@ impl<'ctx, 'src> VineParser<'ctx, 'src> {
 
   fn _parse_expr_prefix(&mut self, span: usize) -> Parse<'src, ExprKind> {
     if self.eat(Token::Return)? {
-      return Ok(ExprKind::Return(Box::new(self.parse_expr_bp(BP::ControlFlow)?)));
+      if self.eat(Token::Semi)? {
```
tjjfvi, line 251:
```suggestion
      if self.check(Token::Semi) {
```

this will leave the `Token::Semi` to be consumed by `parse_stmt`; that way `(return; + 5)` is an error

vine/src/compile/build_stages.rs
stevenhuyn: woah didn't expect that to work. Right now is it undefined behaviour to assign the result of function that returns nothing to a var?

tjjfvi: no, it's just that uninitialized variables are represented with an eraser node, which is the same as an empty tuple

vine/src/parser.rs
stevenhuyn: woops was meant to be check

tjjfvi (approved):

vine/src/compile/build_stages.rs
tjjfvi: or, one can think of it as every variable being initialized as an empty tuple

PR #74, opened 2024-10-24
ivy: improve net evaluation order

tjjfvi:
This causes interactions with the root to happen before the active pairs for global nets. Not sure why this wasn't being done before...


vjjft:
Why is it being done now?

tjjfvi:
Because it improves the statistics of literally every test case

tjjfvi:
though come to think of it I don't know why it would affect allocated/free...

tjjfvi:
ah, that's presumably from ext-fn swapping

Issue #54, opened 2024-10-21
ivm: rework ext system


tjjfvi:
- support more complex ext fns: 
  - take more than two parameters
  - take only one parameter
  - return multiple values
- support extensions via dylibs
- ref-counted ext values

Issue #176, opened 2025-02-12
vine: trait-based operator overloading


tjjfvi:
depends on #2 (to avoid regressing performance)

PR #182, opened 2025-02-15
ivm: custom extrinsics

FranchuFranchu:
This is a PR that adds support for custom extrinsic functions and types to the IVM (#54 ). 
It does not add reference-counted types that are automatically duplicated, or functions with more than 2 arguments. It also does not add loading from dylibs. However, it adds the necessary infrastructure in IVM to support those in the future.


cli/src/common.rs
```diff
@@ -27,19 +27,24 @@ pub struct RunArgs {
...
-    let mut ivm = IVM::new(&heap);
-    ivm.boot(main);
-    if self.workers > 0 {
-      ivm.normalize_parallel(self.workers)
-    } else {
-      ivm.normalize();
-    }
-    if !self.no_stats {
-      eprintln!("{}", ivm.stats);
+    {
```
tjjfvi, line 30: this extra block can be removed

ivm/src/allocator.rs
```diff
@@ -36,7 +36,7 @@ impl<'ivm> Allocator<'ivm> {
   }
 }
 
-impl<'ivm> IVM<'ivm> {
+impl<'ext, 'ivm> IVM<'ext, 'ivm> {
```
tjjfvi, line 39: minor nit, but I'd prefer if it was ordered as `'ivm, 'ext` (a global find & replace should suffice)

ivm/src/ext.rs
```diff
@@ -5,13 +5,225 @@
...
+  n32_ext_ty: Option<ExtTy<'ivm>>,
+  io_ext_ty: Option<ExtTy<'ivm>>,
+
+  phantom: PhantomData<fn(&'ivm ()) -> &'ivm ()>,
+}
+
+impl<'ivm> Extrinsics<'ivm> {
+  pub const MAX_EXT_FN_KIND_COUNT: usize = 0x8FFF;
+  pub const MAX_LIGHT_EXT_TY_COUNT: usize = 0x8FFF;
+  pub const MAX_RC_EX_TTY_COUNT: usize = 0x8FFF;
```
tjjfvi, line 38: these should be `0x7FFF`

ivm/src/ext.rs
```diff
@@ -5,13 +5,225 @@
...
+
+  pub fn register_ext_fn(
+    &mut self,
+    f: impl Fn([ExtVal<'ivm>; 2]) -> [ExtVal<'ivm>; 1] + Sync + 'ivm,
+  ) -> Option<ExtFn<'ivm>> {
```
tjjfvi, line 43: I don't think there's much useful to do if this returns `None`, so all callers are likely to unwrap this. Can this panic, instead of returning an option?

tjjfvi: same for `register_light_ext_ty` below

ivm/src/ext.rs
```diff
@@ -5,13 +5,225 @@
...
+    (ext_fn_map, ext_ty_map)
+  }
+  pub fn call(
+    &self,
+    ext_fn: ExtFn<'ivm>,
+    mut arg0: ExtVal<'ivm>,
+    mut arg1: ExtVal<'ivm>,
+  ) -> ExtVal<'ivm> {
+    let closure =
+      self.ext_fns.get(ext_fn.kind() as usize).expect("ext_fn with given ID not found!");
```
tjjfvi, line 214: in theory, if the `ExtFn` is invariant on the `'ivm`, and if we ensure there can only be one `Extrinsics` for a given `'ivm` (not sure if this is upheld), this check can be removed... probably premature optimization, though; doesn't seem to impact performance

ivm/src/ext.rs
```diff
@@ -5,13 +5,225 @@
...
+    if self.light_ext_ty as usize >= Self::MAX_LIGHT_EXT_TY_COUNT {
+      None
+    } else {
+      let ext_ty = ExtTy::from_id_and_rc(self.light_ext_ty, false);
+      self.light_ext_ty += 1;
+      Some(ext_ty)
+    }
+  }
+  /// Some extrinsics are built-in with IVM
+  pub fn register_builtin_extrinsics(
```
tjjfvi, line 62: I think it would be nice if this could be moved out of `ivm` and into `ivy`. Then the intermediate `HashMap` wouldn't be necessary (the code in `ivy` could call methods on the host) and it would be easier to supply alternate implementations of these default extrinsics.

vine/src/distiller/ext.rs
tjjfvi, line 1: rather than defining an enum here, I think it would be a bit simpler to replace `ExtFnKind` with `&'static str` inside the `VIR`

cli/Cargo.toml
```diff
@@ -8,11 +8,15 @@ anyhow = "1.0.87"
...
+vine = { path = "../vine", optional = true }
+vine-lsp = { path = "../lsp", optional = true }
+ivm = { path = "../ivm" }
 vine-util = { path = "../util" }
 
 [lints]
 workspace = true
+
+[features]
+vine = ["dep:vine", "dep:vine-lsp"]
```
tjjfvi, line 21: can you make `vine` a required feature for the `vine` bin target?

tjjfvi (commented):
looking great; just a few notes

ivm/src/ext.rs
FranchuFranchu: I don't think the fact that there is only one `Extrinsics` for a given `'ivm` is upheld by the typesystem right now.

vine/src/distiller/ext.rs
FranchuFranchu: I think it's an important distinction, as it would refer _only_ to extrinsic functions that are used by Vine language features. Having a `&'static str` is more ambiguous, as it could be interpreted as also including extrinsic functions that are, for example, provided by dynamically linked libraries. These would never need to occur in a `Step::ExtFn`, because they would always be hidden behind `inline_ivy` blocks.

tjjfvi: There are a lot more variants in that enum than are used by Vine language features. That step is probably going away soon anyway though once #176 is implemented (as all of the numeric operations will be handled within `std`). It just seems like a lot of code to define an enum that's only ever converted to a string.

FranchuFranchu: Hmm, you're right. #176 is a very good reason to ditch the enum.

tjjfvi:
the files in `tests/programs/fail` and `tests/programs/fmt` are intentionally left unformatted (they should be excluded in the `dprint.json`)

tjjfvi (approved):
looks great; thanks!

PR #187, opened 2025-02-26
Allow the lexer to work on Windows by permitting `\r` between tokens

chalcolith:
Before this change, the lexer would complain about `\r\n` in files checked out on Windows:

```
error \\?\C:\Users\Gordon\Source\Repos\Vine\vine\vine\examples\hello_world.vi:1:1 - lexing error
error \\?\C:\Users\Gordon\Source\Repos\Vine\vine\vine\std\std.vi:1:1 - lexing error
```

This change allows `\r` between tokens in the Ivy and Vine lexers.


tjjfvi (approved):
good catch; thanks

PR #204, opened 2025-04-08
meta: add nix flake build support

corps:
Adds `flake.nix` to project, which allows one to build vine trivially with nix ecosystem.  For example, I can consume the project with this

```flake.nix
{
  inputs = {
    vine.url = "https://github.com/VineLang/vine";
    vine.inputs.nixpkgs.follows = "nixpkgs"; # I can override which other inputs and build inputs are used
  };
  { vine, ... }:
     # add vine to my system, shell, or as another build input
}
```

I can use `nix flake update` to pull latest vine versions and benefit from binary caching on all the rust components and libraries as well. 

Similarly, anyone with nix could run any "version" of vine by pointing `nix run` at it:

```bash
nix run git+https://github.com/VineLang/vine # with some branch or sha ref, for instance
```

Motivations:
1.  Because vine is unversioned currently and has no other release process.  I'm already tinkering with vine on my laptop and in docker containers.  Swapping out versions with a `flake.lock` is easy while experimenting.
2.  Binary caching in nix is great, saving a great deal of downloading and allowing swapping versions of components trivially.

Downsides:
1.  Dependencies in `Cargo.toml` that do not have checksums need she-256 explicitly in the flake.nix.  Currently this only applies to `class` module.  Either we could move to putting checksum in the `Cargo.toml` like other deps, or just live with the checksum in the `flake.nix`
2.  Some overhead ensuring changes don't break the flake.nix.  Not likely now as long as the build process is mostly rust, nix takes care of this effortlessly.  Willing to contribute time on maintenance.  Possible to easily add check to GitHub actions CI but decided against since it isn't critical for now.  Better that things can break freely.

In support of the ability to build with nix, also added ability to specify VINE_STD_LIB in the environment.  I decided for both a compile time and runtime environment check, but open to suggestion.

### DIFF
--- a/cli/src/vine_cli.rs
+++ b/cli/src/vine_cli.rs
@@ -1,6 +1,5 @@
 use std::{
-  fs,
-  env,
+  env, fs,
   io::{stdin, Read},
   path::PathBuf,
   process::exit,
@@ -92,7 +91,7 @@ fn std_path() -> PathBuf {
     Some(std_path) => {
       path.push(std_path);
       path.push("std.vi");
-    },
+    }
     None => {
       path.push(env!("CARGO_MANIFEST_DIR"));
       path.push("../vine/std/std.vi");
--- a/cli/src/vine_cli.rs
+++ b/cli/src/vine_cli.rs
@@ -87,7 +87,7 @@ fn std_path() -> PathBuf {
   let compile_time_std_path = option_env!("VINE_STD_PATH");
   let runtime_std_path = env::var("VINE_STD_PATH").ok();
 
-  match runtime_std_path.or(compile_time_std_path.map(String::from)) {
+  match runtime_std_path.as_deref().or(compile_time_std_path) {
     Some(std_path) => {
       path.push(std_path);
       path.push("std.vi");
--- a/cli/src/vine_cli.rs
+++ b/cli/src/vine_cli.rs
@@ -1,5 +1,6 @@
 use std::{
   fs,
+  env,
   io::{stdin, Read},
   path::PathBuf,
   process::exit,
@@ -84,8 +85,19 @@ impl CompileArgs {
 
 fn std_path() -> PathBuf {
   let mut path = PathBuf::new();
-  path.push(env!("CARGO_MANIFEST_DIR"));
-  path.push("../vine/std/std.vi");
+  let compile_time_std_path = option_env!("VINE_STD_PATH");
+  let runtime_std_path = env::var("VINE_STD_PATH").ok();
+
+  match runtime_std_path.or(compile_time_std_path.map(String::from)) {
+    Some(std_path) => {
+      path.push(std_path);
+      path.push("std.vi");
+    },
+    None => {
+      path.push(env!("CARGO_MANIFEST_DIR"));
+      path.push("../vine/std/std.vi");
+    }
+  }
   path
 }
 
--- a/flake.nix
+++ b/flake.nix
@@ -8,42 +8,49 @@
     };
   };
 
-  outputs = { nixpkgs, rust-overlay, ... }: 
-  let
-    supportedSystems = [
-      "x86_64-linux"
-      "aarch64-linux"
-      "x86_64-darwin"
-      "aarch64-darwin"
-    ];
-    forAllSystems = nixpkgs.lib.genAttrs supportedSystems;
-    pkgs = system: (import nixpkgs { 
-      inherit system; 
-      overlays = [ rust-overlay.overlays.default ]; 
-    }); 
-  in
-  {
-    packages = forAllSystems (system: with (pkgs system);
-    let 
-      rustPlatform = makeRustPlatform rec {
-        cargo = rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
-        rustc = cargo;
-      };
+  outputs =
+    { nixpkgs, rust-overlay, ... }:
+    let
+      supportedSystems = [
+        "x86_64-linux"
+        "aarch64-linux"
+        "x86_64-darwin"
+        "aarch64-darwin"
+      ];
+      forAllSystems = nixpkgs.lib.genAttrs supportedSystems;
+      pkgs =
+        system:
+        (import nixpkgs {
+          inherit system;
+          overlays = [ rust-overlay.overlays.default ];
+        });
     in
     {
-      default = rustPlatform.buildRustPackage rec {
-        src = lib.cleanSource ./.;
-        pname = "vine";
-        version = "0.1";
-        CFG_RELEASE_CHANNEL = "nightly";
-        VINE_STD_PATH = "${src}/vine/std";
-        cargoLock = {
-          lockFile = "${src}/Cargo.lock";
-          outputHashes = {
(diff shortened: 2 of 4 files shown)


tjjfvi:
@enricozb could you review?

tjjfvi:
> Dependencies in `Cargo.toml` that do not have checksums need she-256 explicitly in the `flake.nix`. Currently this only applies to `class` module.

Hmm, the cargo lock has the full commit hash; is that not enough? I should probably just publish it to crates.io then. Though looks like someone has squatted the name since when I wrote that lib...

cli/src/vine_cli.rs
```diff
@@ -84,8 +84,19 @@ impl CompileArgs {
 
 fn std_path() -> PathBuf {
   let mut path = PathBuf::new();
-  path.push(env!("CARGO_MANIFEST_DIR"));
-  path.push("../vine/std/std.vi");
+  let compile_time_std_path = option_env!("VINE_STD_PATH");
+  let runtime_std_path = env::var("VINE_STD_PATH").ok();
+
+  match runtime_std_path.or(compile_time_std_path.map(String::from)) {
```
tjjfvi, line 90: very minor, but this avoids an allocation :)
```suggestion
  match runtime_std_path.as_ref().or(compile_time_std_path) {
```

corps: Good call!  Adjusted.

corps:
> > Dependencies in `Cargo.toml` that do not have checksums need she-256 explicitly in the `flake.nix`. Currently this only applies to `class` module.
> 
> Hmm, the cargo lock has the full commit hash; is that not enough? I should probably just publish it to crates.io then. Though looks like someone has squatted the name since when I wrote that lib...

Yeah, intuitively the commit hash should be good enough as an immutable content reference, but I think in practice the sha / checksum ends up being needed when verifying integrity of downloads / caches.

It probably isn't a huge deal even with the sha -- if you update your dependency, I can fix the flake.  It basically gives you the updated sha when rebuilding and it doesn't match, so it's a copy paste at worst.

enricozb:
@tjjfvi @corps 
> Hmm, the cargo lock has the full commit hash; is that not enough? I should probably just publish it to crates.io then. Though looks like someone has squatted the name since when I wrote that lib...

I believe you need sha's in the nix files that package rust crates/binaries _only when dependencies are unpublished_. I was packaging a project recently and ran into this when using git dependencies in my `Cargo.toml`.

flake.nix
```diff
@@ -0,0 +1,61 @@
...
+    devShells = forAllSystems (system: with (pkgs system);
+      {
+        default = mkShell {
+          buildInputs = [ rustup ];
+          shellHook = ''
+            source <(rustup completions bash)
+            source <(rustup completions bash cargo)
+          '';
+        };
+      }
```
enricozb, line 35: I think this is fine if one uses rustup, not necessarily the standard configuration though.

flake.nix
```diff
@@ -0,0 +1,61 @@
...
+    );
+
+    packages = forAllSystems (system: with (pkgs system);
+    let 
+      rustPlatform = makeRustPlatform rec {
+        cargo = rust-bin.nightly."2024-12-24".default;
```
enricozb, line 41: can we infer this from the toolchain file?

flake.nix
corps: Fair.  Is there a specific preference?  I used this because it was easy to setup without any other strong opinions.

flake.nix
corps: Added, thank you.

enricozb:
@corps how affected are you if i remove the `defaultShells` stanza? Trying to make this flake focus on packaging rather than dev.

corps:
> @corps how affected are you if i remove the `defaultShells` stanza? Trying to make this flake focus on packaging rather than dev.

Not terribly.  I thought of it more as a stub to expand on later if it was helpful.

corps:
@enricozb I went ahead and just removed the stanza from my forked branch.

flake.nix
```diff
@@ -0,0 +1,49 @@
...
+        cargoLock = {
+          lockFile = "${src}/Cargo.lock";
+          outputHashes = {
+            "class-0.1.0" = "sha256-ye8DqeDRXsNpTWpGGlvWxSSc1AiXOLud99dHpB/VhZg=";
+          };
+        };
+      };
+    });
+  };
+}
```
enricozb, line 49: Thoughts on this slight minimization using `flake-utils`?

```suggestion
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
    flake-utils.url = "github:numtide/flake-utils";
    rust-overlay = {
      url = "github:oxalica/rust-overlay";
      inputs.nixpkgs.follows = "nixpkgs";
    };
  };

  outputs = { nixpkgs, flake-utils, rust-overlay, ... }:
    flake-utils.lib.eachDefaultSystem (system:
      let
        overlays = [ (import rust-overlay) ];
        pkgs = import nixpkgs { inherit system overlays; };
        rust = pkgs.makeRustPlatform rec {
          cargo = pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
          rustc = cargo;
        };
      in {
        packages = {
          default = rust.buildRustPackage rec {
            src = pkgs.lib.cleanSource ./.;
            pname = "vine";
            version = "0.1.0";
            CFG_RELEASE_CHANNEL = "nightly";
            VINE_STD_PATH = "${src}/vine/std";
            cargoLock = {
              lockFile = "${src}/Cargo.lock";
              outputHashes = {
                "class-0.1.0" = "sha256-ye8DqeDRXsNpTWpGGlvWxSSc1AiXOLud99dHpB/VhZg=";
              };
            };
          };
        };
      });
}
```

enricozb:
Also I'm trying to understand the rust build system a bit better, but it's not clear to me what `CFG_RELEASE_CHANNEL` does at build-time. What's the purpose of it here?

flake.nix
corps: My personal experience,

is that flake utils is great if you're using a lot of its features, but it seems overkill merely for the system loop.  Without using flake utils, there is a 12 line difference, but there's also one less dependency to manage (one less thing to break) because we just use nixpkgs standard tooling, and furthermore, supported systems is more explicit (rather than implicit).

I won't cry either way, but if it's merely my opinion, I'd wait for more reasons to use flake utils than throw it in immediately.

corps:
> Also I'm trying to understand the rust build system a bit better, but it's not clear to me what `CFG_RELEASE_CHANNEL` does at build-time. What's the purpose of it here?

CFG_RELEASE_CHANNEL is not well documented, but it is used by some tools to conditionalize features and behavior.  I'll be honest, I need to double check -- I first tried building vine without this and using standard toolchain, but some dependency was complaining it wouldn't compile until it saw a nightly release channel.  When I have some more time later I can double check and provide more context on that.

enricozb:
Hmm it built for me without that flag, let me know if it's still broken for you.

w.r.t flake-utils, i think i can buy your argument. i'd format the flake with some formatter (to clear trailing spaces) and i think i'd be comfortable approving

enricozb (approved):

PR #207, opened 2025-04-22
std: add Iterator trait

nilscrm:
Adds an `Iterator` trait to the standard library to unify iterators over collections.

This will also enable common operations on iterations such as `zip` or `chain` and lays the ground work for an eventual support in for loops.


tjjfvi:
The `Iterator` trait looks reasonable to me. I'd prefer not to have an `IntoIterator` trait at the moment though – not being able to ensure it actually creates an iterator is a bit of an issue, and I'd like to give more thought to what the API should look like.

nilscrm:
> The `Iterator` trait looks reasonable to me. I'd prefer not to have an `IntoIterator` trait at the moment though – not being able to ensure it actually creates an iterator is a bit of an issue, and I'd like to give more thought to what the API should look like.

Fair, I guess using the `IntoIterator` trait only makes sense in situations where you can actually rely on the fact that the `Iter` type is actually an iterator anyway.

vine/std/data/List.vi
```diff
@@ -133,27 +134,29 @@ pub mod List {
     self = output;
   }
 
-  pub struct Iter[T](N32, &Buf[T]);
+  struct Iter[T](N32, &Buf[T]);
```
tjjfvi, line 137:
```suggestion
  pub struct Iter[T](N32, &Buf[T]);
```

nilscrm: What's the benefit of making this struct public? The user should only be able to create an `Iter` through the `.iter()` method and use it through the `Iterator` interface, right?

tjjfvi: Yes, but making the type public allows it to be stored in other iterator structs etc.

tjjfvi:
Also, looks like some of the implementations are named `iter` and some are named `iterator`; I don't have a preference, but it should be consistent.

tjjfvi (approved):

PR #208, opened 2025-04-23
ivm: add depth statistic

nilscrm:
Adds a `--breath-first` flag to reduce the net breath first in order to determine the depth of the computations which helps understand the parallelism on the computation.


ivm/src/ivm.rs
```diff
@@ -82,4 +83,25 @@ impl<'ivm, 'ext> IVM<'ivm, 'ext> {
       self.interact(a, b)
     }
   }
+
+  /// Normalize all nets in breadth-first traversal.
+  ///
+  /// This is useful to get the depth (longest critical path) of the computation
+  /// to understand the parallelism of the program.
+  pub fn normalize_breadth_first(&mut self) {
```
nilscrm, line 91: I'm not sure I'm 100% happy with the name, can you think of something better?

ivm/src/stats.rs
```diff
@@ -71,6 +74,12 @@ impl Display for Stats {
       ("", None),
     ]
     .into_iter()
+    .chain(
+      (self.depth != 0)
+        .then_some(())
+        .into_iter()
+        .flat_map(|_| [("Computation Depth", Some((self.depth, ""))), ("", None)]),
```
nilscrm, line 81: Should these somewhat convoluted expression with `Option`s maybe be just turned into a single `if` statement and a `chain`? (same for the other `chain`s)

nilscrm (commented):
Should I maybe add some tests for this as well?

ivm/src/ivm.rs
```diff
@@ -82,4 +83,25 @@ impl<'ivm, 'ext> IVM<'ivm, 'ext> {
...
+    let start = Instant::now();
+    let mut work;
+    loop {
+      work = mem::take(&mut self.active_fast);
```
tjjfvi, line 95: Using `mem::swap` instead of `mem::take` here will allow reuse of the buffer.

nilscrm: `mem::swap` currently does not work as the for loop takes ownership of the buffer (the interact methods needs ownership of the ports). Is there a way to change this?

tjjfvi: Ah, you can use `work.drain(..)` for the for loop.

nilscrm: Perfect

tjjfvi:
And yeah, some tests would be good. Perhaps you can add a boolean param to `test_vi` which tests it in `--breadth-first` mode. IMO that can/should be added to any tests where it is no too slow to do so, as it would be an interesting statistic to see.

ivm/src/stats.rs
tjjfvi: On this topic, I think this stats should be formatted as:
```rs
Interactions
  Total                890_255_524
  Depth                    275_843
  Annihilate           409_360_049
  Commute                  169_830
  Copy                  99_659_432
  Erase                 95_452_421
  Expand                62_238_264
  Call                 162_309_108
  Branch                61_066_420

Memory
  Heap                  63_350_064 B
  Allocated         18_388_838_448 B
  Freed             18_388_838_448 B

Performance
  Time                      32_096 ms
  Speed                 27_736_505 IPS
```
I.e. have `Depth` in the `Interactions` section under `Total`

nilscrm: Ah yeah makes sense, totally forgot this section had a `total` as well for a sec.

tjjfvi (approved):

PR #220, opened 2025-04-28
WIP (Adding IO Operations and boxed exts)

corps:
Apologies for the long delay, I have been traveling lately and haven't had a lot of time to put into this, but I wanted to open a PR to share the status and get some feedback before I put more time into it. 

Two things,

1.  I borrowed and integrated @FranchuFranchu 's incomplete Arc work, and I am running into an issue related to packing pointers into `ExtVal`:

```rust
  #[inline(always)]
  pub unsafe fn new_rc_raw(ty: ExtTy<'ivm>, pointer: *const ()) -> Self {
    assert!(ty.is_rc());
    unsafe {
      Self::from_word(Word::from_ptr(pointer).map_bits(|payload| {
        let payload = pointer.addr() as u64;
        // Ensure the payload pointer does not conflict with any other field
        assert!(payload & 0xFF000007 == 0);
        ty.0 as (u64) << 48 | payload | Tag::ExtVal as u64
      }))
    }
  }
```

The issue is that the pointer bits definitely conflict with ExtVal's type and tags (hitting the assert), and I'm not sure how to proceed.  I have some silly hacks but are likely not the intention, so I'm curious.  Is the idea to possibly be using the IVM's heap to solve this similar to how wires and ports are managed or?

2.  The IO interface in general definitely feels brittle.  There's some asymmetry with how IO works and how Files work (for instance, there isn't a good way right now to convert N32 bytes to Char, so you end up with two interfaces for dealing with input data, which isn't great), and there's definitely code duplication in wiring IO around and replicating similar effects as File does.

I had half an idea to generalize to an FFI interface instead of writing `File` the way it is.  In an FFI interface world, the vine level ext functions would just expose ways to invoke external shared library methods using general `Buffer` types, and the Ivy side would just implement the serialization and deserialization of buffers and invoking these methods.  FFI would have a unified interface through IO that can share an easy to understand model.

  One advantage of this is that it would allow me to quickly experiment with some of my project ideas without altering the core language in an indefensible way (how I decide to implement files and networks are just experiments on the FFI layer, which can be discarded or generalized in the future).


tjjfvi:
With regard to this assert:
```rs
        assert!(payload & 0xFF000007 == 0);
```
I believe that should be:
```rs
        assert!(payload & 0xFFFF_0000_0000_0007 == 0);
```

tjjfvi:
With regard to having a general FFI interface (gotta love good ol' RAS syndrome): the idea behind the extrinsic system is that it should ultimately be extensible by dynamic libraries (as well as much easier to use) so that people can write application-specific extensions. I don't want to be the gatekeeper of all IVM IO :). That just hasn't been implemented yet.

A buffer-based system could be done, but I think it would be less ideal. Though certainly once the extrinsic is extensible some buffer-type thing could be built on top of it if that was more convenient for some application.

corps:
Thank you for the context, that all makes more sense then.
I'll update again when I've some time to put back into the branch.

corps:
Just a ping that I'm still traveling, but intend to return to and clean this up.

I have one question though while I was on the train recently,

> the idea behind the extrinsic system is that it should ultimately be extensible by dynamic libraries (as well as much easier to use) so that people can write application-specific extensions. I don't want to be the gatekeeper of all IVM IO :). That just hasn't been implemented yet.

This is super ideal!  Are you imagining dynamic library loading or static compilation?  Actually, I'm more interested in helping with that idea because it would help me with some of the projects I had in mind.

tjjfvi:
Well, both – there should be some reasonable API so that someone could make their own CLI that wraps IVM with some extrinsics, but I'd also like to support dynamic loading, as I think that will be a lot easier to use.

It actually works out well that you've been traveling... it hopefully gives me time to get fork/drop merged before you're looking back to this, which would be convenient because I think it can be used to simplify this a bit. (Particularly, removing the RC stuff, and letting extvals be ~arbitrary pointers. [Some context on Discord.](https://discord.com/channels/1246152587883970662/1330960700952154213/1366557008257355786))

PR #225, opened 2025-05-13
ivm: add breadth stat

tjjfvi:
Equal to `Total / Depth`.

I've also indented `Depth` and `Breadth` to distinguish them from the interaction types.


nilscrm (approved):
looks good

PR #228, opened 2025-05-20
std: add starts_with method to String

nilscrm:
Currently, like `strip_prefix`, `starts_with` takes the prefix as reference and I can see reasons for doing this especially when reusing a prefix multiple times and repeatedly having diverging prefixes.

However, I can also see reasoning for why is might be nice to have `starts_with` take the prefix as value for performance reasons and better parallelism. Wdyt?


tjjfvi (approved):
Looks good to me. I think reference is good for now and we can reconsider down the road if warranted.

PR #229, opened 2025-05-20
tooling: use installed vine bin for formatting

nilscrm:
Files get formatted instantly rather than in ~1s when using cargo and formatting also continues vine doesn't build while e.g. working on the compiler.

The downside is that changes to the formatter itself require a `cargo install` to propergate but I think that is reasonable and justified by the benefits.


tjjfvi:
Perhaps a middleground would be to use `target/release/vine`? The LSP builds that anyway so it will exist. Removes the overhead of cargo, allows it to work even when the compiler isn't building, but easily changes when moving between branches of the compiler.

tjjfvi (approved):

PR #231, opened 2025-05-22
std: implement Eq for List and String

nilscrm:


tjjfvi:
I think it would make sense to implement Eq for List and then use that for String, like is done with Ord.

tjjfvi (approved):

PR #233, opened 2025-05-22
vine: add exponentiation operator

nilscrm:
* Add right-associative exponentiation operator `**`.
* Add `Pow` trait to std
* Implement exponentiation by `N32` for all numeric types


tjjfvi (changes requested):
`BP` stands for binding power, not `BinPrecedence`. (It's used outside of binary operators.) It's also used so much in the parser that I would prefer it stayed abbreviated.

tjjfvi:
For `F32` exponentiation, it might make sense to add a `f32_pow_n32` extrinsic function that is implemented in Rust. (The other exponentiation methods for it will presumably be as well.)

nilscrm:
> `BP` stands for binding power, not `BinPrecedence`. (It's used outside of binary operators.) It's also used so much in the parser that I would prefer it stayed abbreviated.

I see, added a doc comment for poor people like to know what it means :)

tests/programs/op_precedence.vi
```diff
@@ -0,0 +1,4 @@
+
+pub fn main(&io: &IO) {
+  io.println("{4 + 4 ** 3 ** 2 / 2}");
```
tjjfvi, line 3: For one-off tests like this I'm often inclined to throw them in the `misc.vi` repl test. Not necessary just throwing it out there.

nilscrm:
> For `F32` exponentiation, it might make sense to add a `f32_pow_n32` extrinsic function that is implemented in Rust. (The other exponentiation methods for it will presumably be as well.)

I'm not sure. The only reason rust has an intrinsic for this is because llvm has one for it. For small exponents (–3 … 27) it gets optimized to a few multiplies/divides. For larger exponents it uses the `libm` or rust `compiler_builtins` functions which also just do exponentiation by squaring. That means the exponentiation is never seen by any hardware. If Ivy introduced an extrinsic for this, the hardware would have to deal with it which is not ideal.

nilscrm:
> > For `F32` exponentiation, it might make sense to add a `f32_pow_n32` extrinsic function that is implemented in Rust. (The other exponentiation methods for it will presumably be as well.)
> 
> I'm not sure. The only reason rust has an intrinsic for this is because llvm has one for it. For small exponents (–3 … 27) it gets optimized to a few multiplies/divides. For larger exponents it uses the `libm` or rust `compiler_builtins` functions which also just do exponentiation by squaring. That means the exponentiation is never seen by any hardware. If Ivy introduced an extrinsic for this, the hardware would have to deal with it which is not ideal.

Also looked at what llvm does for small constant exponents and it's literally also just inlining of the exponentiation by squaring algorithms which we can probably get simply by pre-reducing.

tests/snaps/vine/repl/misc.repl.vi
```diff
@@ -146,19 +150,21 @@ io = #io
 
 io = #io
 > 3.as[N64].pow_n32(33).to_string()
```
tjjfvi, line 152: this should be updated

nilscrm: thanks

tjjfvi (approved):

PR #237, opened 2025-05-23
vine: rework structs & enums

tjjfvi:
Structs are now newtype wrappers around a single base type, written `struct NewType(InnerType)`. To have multiple fields, one can use a tuple or object as the base. A struct can be unwrapped with the postfix `!` operator. This is automatically done when accessing fields on a struct.

Struct data can have visibility. Structs can only be constructed, pattern-matched, or unwrapped if their data is visible to the current scope.

Enum variants can now either be dataless or wrap a single data type. Similarly to structs, this data can be a tuple or object to have multiple fields.

To avoid syntactic ambiguity, the `pub(some_ancestor)` syntax has been changed to `pub.some_ancestor`.


nilscrm (commented):
Can you explain the motivation for this a little bit. Looking at the example test programs, I practically always prefer the old version.

Is this more about the language syntax or the internal representation of structs/enum?

tjjfvi:
This isn't really about the syntax but the semantics.

Before, structs did two things: create a nominal type, and encapsulate a set of fields. There were correspondingly two kinds of structs, tuple structs, and object structs, for the two kinds of fields. If you *just* wanted to create a nominal type, without combining multiple fields, you couldn't -- the two behaviors were tied together. The workaround was to create a tuple struct with just one field (the newtype pattern in Rust). Accessing the inner data was then done with `.0` (which is ugly, and not really semantically correct).

Now, structs do *one* thing: create a nominal type. If you want to encapsulate a set of fields, you can use structs in conjunction with one of the existing things that do that -- tuples or objects. The newtype pattern is no longer necessary, as structs are, at their core, newtypes. `String`, for example, is now just a newtype wrapper around `List[Char]`.

This is also important for impl derivation. With impl derivation, the idea is that each struct will have a "base type" that it can derive implementations for.

Before, the base type of a tuple struct would be an anonymous tuple, and the base type of an object struct would be an anonymous object. The issue with this is that the newtype pattern wouldn't quite work correctly -- `struct Foo(Bar)` would have a base type of `(Bar,)`, not `Bar`. One could consider making an exception for this, but it would add to the already long list of weird edge cases with `struct`s. For example, though object structs could be constructed with `Struct(base_type)`, tuple structs could not.

With this change, the story is much simpler. Structs are simply nominal type wrappers. The base type of a struct is simply the type it wraps. Every struct can be constructed with `Struct(base_type)`.

This change simplifies the language model quite a bit. It does make the syntax for tuple structs a little uglier -- but that, IMO, is an acceptable cost.

nilscrm:
I can see how this semantic model makes things nicer. I'm also not a huge fan of `.0`. I will say so that I believe `Struct(base_type)` is a really uncommon construction and you should probably not have a `base_type` as a value in the first place.

I also feel like it should be possible to make this semantic change and not make the syntax uglier. Especially since structs and tuples are everywhere, I think getting the syntax for them right is important.

I could actually see that making the exception that `struct Foo(Bar)` means the base type is `Bar` instead of `(Bar,)` might be reasonable and a lot nicer syntax wise.

tests/snaps/vine/repl/advanced_repl.repl.vi
```diff
@@ -1,22 +1,23 @@
 
 io = #io
 > struct Foo(N32, F32, Char, String);
```
nilscrm, line 3: This struct def needs to be updated.

tests/snaps/vine/repl/misc.repl.vi
nilscrm, line 1: I think the double parens are not needed anymore, right?

tests/snaps/vine/repl/randomness.repl.vi
nilscrm, line 1: Same as above, double parens are probably not needed.

tests/snaps/vine/fail/atypical.txt
```diff
@@ -1,37 +1 @@
...
-error tests/programs/fail/atypical.vi:21:14 - expected type `U`; found `T`
-error tests/programs/fail/atypical.vi:22:7 - expected a complete pattern
-error tests/programs/fail/atypical.vi:23:4 - cannot find impl of trait `Add[String, N32, ?31]`
-error tests/programs/fail/atypical.vi:24:7 - expected type `N32`; found `&?32`
-error tests/programs/fail/atypical.vi:27:10 - expected type `()`; found `F32`
-error tests/programs/fail/atypical.vi:28:14 - cannot find impl of trait `Add[~N32, ~N32, ?41]`
-error tests/programs/fail/atypical.vi:29:3 - cannot find impl of trait `Add[N32, F32, N32]`
-error tests/programs/fail/atypical.vi:30:3 - type `N32` has no method `parse`
-error tests/programs/fail/atypical.vi:6:27 - expected type `()`; found `N32`
+error tests/programs/fail/atypical.vi:38:17 - expected one of {ColonColon, CloseParen, OpenBracket}; found Some(Comma)
```
nilscrm, line 1: Struct def of `Thing` needs to be fixed.

tests/snaps/vine/fail/visibility.txt
```diff
@@ -1,11 +1 @@
...
-error - `::visibility::main` is only visible within `::visibility`
-error tests/programs/fail/visibility.vi:21:24 - `::visibility::lib::y` is only visible within `::visibility::lib`
-error tests/programs/fail/visibility.vi:21:7 - `::visibility::lib::y` is only visible within `::visibility::lib`
-error tests/programs/fail/visibility.vi:21:15 - `::visibility::lib::y` is only visible within `::visibility::lib`
-error tests/programs/fail/visibility.vi:27:8 - `::visibility::lib::a::b` is only visible within `::visibility::lib::a`
-error tests/programs/fail/visibility.vi:28:5 - `::visibility::lib::c` is only visible within `::visibility::lib`
-error tests/programs/fail/visibility.vi:29:8 - `::visibility::lib::d::e` is only visible within `::visibility::lib`
-error tests/programs/fail/visibility.vi:33:17 - circular import
-error tests/programs/fail/visibility.vi:19:3 - type `IO` has no method `read_char`
+error tests/programs/fail/visibility.vi:12:12 - expected one of {ColonColon, And, AndAnd, Tilde, OpenBrace, OpenParen, Hole, Pub, Fn, Ident}; found Some(CloseParen)
```
nilscrm, line 1: A few syntax error need to be fixed.

vine/src/chart.rs
```diff
@@ -143,11 +144,8 @@ pub enum ValueDefKind<'core> {
     ret: Option<Ty<'core>>,
     body: Block<'core>,
   },
-  Ivy {
-    ty: Ty<'core>,
-    net: Net,
-  },
```
nilscrm, line 149: Is this Ivy `ValueDefKind` related to struts and enums?

vine/src/checker.rs
```diff
@@ -405,16 +428,28 @@ impl<'core, 'a> Checker<'core, 'a> {
...
-        let fields = &self.types.adt_types[*adt_id][*variant_id];
-        let adt = Type::Adt(*adt_id, (0..generics.type_params.len()).map(Type::Param).collect());
-        if fields.is_empty() {
-          adt
+      ValueDefKind::Struct(struct_id) => {
+        let struct_def = &self.chart.structs[*struct_id];
+        let generics = &self.chart.generics[struct_def.generics];
+        let data = &self.types.struct_types[*struct_id];
+        Type::Fn(
+          vec![data.clone()],
```
nilscrm, line 436: The old version seemed to have had a special case for when there were no fields. I think now this is only done for enums. Is it okay to remove this behavior for structs?

vine/src/checker/check_pat.rs
```diff
@@ -83,37 +111,4 @@ impl<'core> Checker<'core, '_> {
...
-    form: Form,
-    refutable: bool,
-  ) -> Result<Type<'core>, Diag<'core>> {
-    let type_params = self.check_generics(generics, self.chart.adts[adt].generics, true);
-    let field_tys = self.types.adt_types[adt][variant]
-      .iter()
-      .map(|t| t.instantiate(&type_params))
-      .collect::<Vec<_>>();
-    if fields.len() != field_tys.len() {
-      Err(Diag::BadFieldCount {
```
nilscrm, line 104: `BadFieldCount` is never reported now. Maybe this needs to be checked for tuple and/or objects.

nilscrm (changes requested):

nilscrm:
> I've made it so that for constructors, `Foo(a, b, ...)` is sugar for `Foo((a, b, ...))`. This removes almost all cases of the double-paren form, whilst leaving the semantics intact.

Nice!

> (Double parens are still required for declaring the struct, like `struct Foo((N32, F32))`, but this is because the visibility modifier goes between those parens, like `struct Foo(pub (N32, F32))`.)

Do you think one could still introduce syntactic sugar for it? In cases where one wants to explicitly annotate the visibility one can always fall back to the more verbose syntax (or potentially we could come up with a better syntax for the visibility `struct Foo(pub (N32, F32))` doesn't look particularly nice)

> This also means that `Some((a, b))` can be written as `Some(a, b)`, which is something I've always wanted in Rust.

Nice, would this also work to write `Ok()` instead of `Ok(())` which is something I've always wanted in Rust.

> @nilscrm Does this address your concerns?

Yep, definitely an improvement!

vine/src/checker.rs
tjjfvi: Yeah; structs now always wrap data, but for enums it's optional.

vine/src/chart.rs
nilscrm: Ah I see, would be cleaner to do it in a separate PR but I guess it's fine as I imagine it being a bit of a hassle to split the PR now.

tjjfvi:
> Do you think one could still introduce syntactic sugar for it? In cases where one wants to explicitly annotate the visibility one can always fall back to the more verbose syntax (or potentially we could come up with a better syntax for the visibility `struct Foo(pub (N32, F32))` doesn't look particularly nice)

I don't think syntax sugar should be introduced there. It makes sense for the expressions/patterns because they're used several orders of magnitude more, and there isn't salient information (the `pub`) to put there. It's good for the syntax for declaring the struct to align very closely to the semantics, IMO.

> Nice, would this also work to write `Ok()` instead of `Ok(())` which is something I've always wanted in Rust.

Yeah, this works as well.

nilscrm:
> It's good for the syntax for declaring the struct to align very closely to the semantics, IMO.

I feel like there is also an argument to closely align the the syntax of declaring a struct to the syntax of using a struct.

> > Nice, would this also work to write `Ok()` instead of `Ok(())` which is something I've always wanted in Rust.
> 
> Yeah, this works as well.

Nice

nilscrm:
What do you think of the syntax `struct Foo pub(N32, F32)`

nilscrm (approved):

PR #238, opened 2025-05-23
vine: replace primitive types with opaque types

tjjfvi:
Just affects the way `N32` etc. are defined in the standard library. Makes the compiler a bit simpler. Also means that other opaque types can easily be defined e.g. an opaque `File` for #220.

### DIFF
--- a/vine/src/checker/check_expr.rs
+++ b/vine/src/checker/check_expr.rs
@@ -2,7 +2,7 @@ use std::{collections::BTreeMap, mem::take};
 
 use crate::{
   ast::{Expr, ExprKind, Generics, ImplKind, Label, Span},
-  chart::{Builtins, TypeDefId, VariantId},
+  chart::{TypeDefId, VariantId},
   checker::{Checker, Form, Type},
   diag::Diag,
 };
@@ -363,7 +363,7 @@ impl<'core> Checker<'core, '_> {
         ty
       }
       ExprKind::Call(func, args) => self.check_call(func, args, span),
-      ExprKind::Bool(_) => self.builtin_ty(span, "Bool", |b| b.bool),
+      ExprKind::Bool(_) => self.bool(span),
       ExprKind::Not(inner) => {
         let ty = self.check_expr_form(inner, Form::Value);
         let Some(op_fn) = self.chart.builtins.not else {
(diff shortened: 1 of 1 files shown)


vine/src/checker.rs
```diff
@@ -615,20 +618,15 @@ new_idx!(pub Var);
...
-  IO,
   Tuple(Vec<Type<'core>>),
   Object(BTreeMap<Ident<'core>, Type<'core>>),
   Fn(Vec<Type<'core>>, Box<Type<'core>>),
   Ref(Box<Type<'core>>),
   Inverse(Box<Type<'core>>),
   Adt(AdtId, Vec<Type<'core>>),
   Trait(TraitDefId, Vec<Type<'core>>),
-  Opaque(usize),
+  Opaque(TypeDefId, Vec<Type<'core>>),
```
nilscrm, line 628: It's not immediately clear that this vector of types are generic parameter. Potentially we could add explicit field names (or a comment).

vine/src/checker/check_expr.rs
```diff
@@ -613,4 +616,21 @@ impl<'core> Checker<'core, '_> {
     }
     ty
   }
+
+  fn builtin_ty(
+    &mut self,
+    span: Span,
+    name: &'static str,
+    f: fn(&Builtins) -> Option<TypeDefId>,
```
nilscrm, line 624: `f` is called exactly once with a single argument and so I think the function could just as well directly get the `Option<TypeDefId>` making the function and call sites cleaner.

nilscrm (changes requested):

vine/src/checker/check_expr.rs
tjjfvi: each call site needs a different field:
```rs
      ExprKind::N32(_) => self.builtin_ty(span, "N32", |b| b.n32),
      ExprKind::I32(_) => self.builtin_ty(span, "I32", |b| b.i32),
      ExprKind::Char(_) => self.builtin_ty(span, "Char", |b| b.char),
      ExprKind::F32(_) => self.builtin_ty(span, "F32", |b| b.f32),
```

nilscrm: What I mean is changing it to
```rust
      ExprKind::N32(_) => self.builtin_ty(span, "N32", b.n32),
      ExprKind::I32(_) => self.builtin_ty(span, "I32", b.i32),
      ExprKind::Char(_) => self.builtin_ty(span, "Char", b.char),
      ExprKind::F32(_) => self.builtin_ty(span, "F32", b.f32),
```

vine/src/checker.rs
tjjfvi: Named fields / better documentation is needed across the compiler. I'm inclined to leave this consistent with the rest of the variants for now and improve this later across the board.

tjjfvi: Yeah, it's definitely sorely needed. I'll plan to do this after I finish up fork/drop.

PR #242, opened 2025-05-25
tooling: fix gitignore notation

nilscrm:
Git actually doesn't understand the bracket notation so split this into two lines.


tjjfvi (approved):
Thanks; jj accepts it so I didn't notice lol

PR #244, opened 2025-05-25
std: fix iterator dropping in List eq

nilscrm:
When lists or strings of the same length were compared that are not equal, not dropping the iterator deletes part of the string.


tjjfvi (approved):
fork/drop can't come quickly enough...

PR #250, opened 2025-05-27
vine: add `Fork` and `Drop` traits [fork/drop]

tjjfvi:
Adds the `Fork` and `Drop` traits to the standard library, and implements them for all relevant types. Note that the compiler does not yet automatically call them.

I renamed the `Iterator::drop` method to `Iterator::drop_iter` to avoid a naming conflict for now. Ultimately this method will be removed and its role taken by `drop`, but I would like to leave it separate for now for simplicity / compatibility.

### DIFF
--- a/vine/std/ops/flex.vi
+++ b/vine/std/ops/flex.vi
@@ -1,0 +1,98 @@
+
+pub trait Fork[T] {
+  fn .fork(&self: &T) -> T;
+}
+
+pub trait Drop[T] {
+  fn .drop(self: T);
+}
+
+pub mod Fork {
+  pub fn copy[T](&value: &T) -> T {
+    inline_ivy! (x <- value, x1 -> value) -> T {
+      x2
+      x = copy(x1 x2)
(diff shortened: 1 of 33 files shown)


vine/std/ops/flex.vi
```diff
@@ -0,0 +1,100 @@
...
+  fn .drop(self: T);
+}
+
+pub mod Fork {
+  pub fn copy[T](&value: &T) -> T {
+    inline_ivy! (x <- value, x1 -> value) -> T {
+      x2
+      x = copy(x1 x2)
+    }
+  }
```
nilscrm, line 18: This operation is not safe for all Types that implement `Fork` right? Should we at least add some warning/safety comment?

nilscrm (commented):
Nice, Fork/Drop is really getting closer.

For the `Iterator::drop_iter()` methods, couldn't they just be `Drop` implementations?

nilscrm (approved):

PR #261, opened 2025-05-31
vine: add ranges

nilscrm:
Add support for range notation in Vine (`a..b` and `a..=b`).

Defines a generic range type in the standard library than can be used flexibly and is used by the added `slice` method on lists.


vine/std/data/List.vi
```diff
@@ -32,17 +32,20 @@ pub mod List {
...
+
+    while !range.contains(i) && self.pop_front() is Some(_) {
+      i += 1;
+    }
+    while self.pop_front() is Some(value) && range.contains(i) {
+      slice.push_back(value);
+      i += 1;
     }
+    slice
   }
```
tjjfvi, line 49: Can save some redundant checks with
```
  pub fn .slice[T, B1, B2; Bound[B1, N32], Bound[B2, N32]](self: List[T], Range[B1, B2](start, end)) -> List[T] {
    let len = self.len();
    let i = 0;
    let slice = [];

    while !start.left_of(i) && self.pop_front() is Some(_) {
      i += 1;
    }
    if end.right_of(len - 1) {
      return self;
    }
    while end.right_of(i) && self.pop_front() is Some(value) {
      slice.push_back(value);
      i += 1;
    }
    slice
  }
```

vine/src/parser.rs
```diff
@@ -150,6 +150,10 @@ impl<'core, 'src> VineParser<'core, 'src> {
           "ge" => Builtin::ComparisonOp(ComparisonOp::Ge),
           "Fork" => Builtin::Fork,
           "Drop" => Builtin::Drop,
+          "Range" => Builtin::Range,
+          "Unbounded" => Builtin::RangeBound(Bound::Unbounded),
+          "BoundInclusive" => Builtin::RangeBound(Bound::Inclusive),
+          "BoundExclusive" => Builtin::RangeBound(Bound::Exclusive),
```
tjjfvi, line 156: I think these should either all be prefixed with `Bound` or all not

vine/src/chart.rs
```diff
@@ -51,6 +51,11 @@ pub struct Builtins {
 
   pub fork: Option<TraitDefId>,
   pub drop: Option<TraitDefId>,
+
+  pub range: Option<StructId>,
+  pub bound_exclusive: Option<StructId>,
+  pub bound_inclusive: Option<StructId>,
+  pub unbounded: Option<StructId>,
```
tjjfvi, line 58: Similarly I think these should either all be `bound_` or all not

vine/src/ast.rs
```diff
@@ -584,6 +590,13 @@ pub enum ComparisonOp {
   Ge,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Bound {
```
tjjfvi, line 594: Since this isn't used outside of the `Builtins` enum, I think this should be inlined there (so there would be `Builtins::BoundUnbounded`,` `Builtins::BoundInclusive`, `Builtins::BoundExclusive`)

vine/std/ops/range.vi
```diff
@@ -0,0 +1,58 @@
...
+  fn .right_of(self: &B, value: T) -> Bool;
+}
+
+pub mod Bound {
+  #[builtin = "BoundUnbounded"]
+  pub struct Unbounded(());
+  #[builtin = "BoundInclusive"]
+  pub struct Inclusive[T](T);
+  #[builtin = "BoundExclusive"]
+  pub struct Exclusive[T](T);
```
tjjfvi, line 15:
```suggestion
  #[builtin = "BoundUnbounded"]
  pub struct Unbounded(pub ());
  #[builtin = "BoundInclusive"]
  pub struct Inclusive[T](pub T);
  #[builtin = "BoundExclusive"]
  pub struct Exclusive[T](pub T);
```

vine/std/ops/range.vi
```diff
@@ -0,0 +1,58 @@
+
+use comparison::Ord;
+
+pub trait Bound[B, T] {
+  fn .left_of(self: &B, value: T) -> Bool;
+  fn .right_of(self: &B, value: T) -> Bool;
```
tjjfvi, line 6: These should take `value` by reference (like `Ord`)

vine/std/ops/range.vi
```diff
@@ -0,0 +1,58 @@
...
+      }
+    }
+  }
+}
+
+#[builtin = "Range"]
+pub struct Range[B1, B2]((B1, B2));
+
+pub mod Range {
+  pub fn .contains[B1, B2, T; Bound[B1, T], Bound[B2, T]](&Range[B1, B2](start, end), value: T) -> Bool {
```
tjjfvi, line 55: This should also be by reference

vine/std/ops/range.vi
nilscrm: Should the `Range` tuple be public as well then?

tjjfvi: yeah

tjjfvi (approved):

PR #272, opened 2025-06-08
vine: disallow marking impl fns as methods

tjjfvi:
Currently, impl fns can be marked as methods, but this does not really affect anything; it is whether the trait fn is marked as a method that matters. This requires that impl fns not be marked as methods (much like they cannot have visibility).

### DIFF
--- a/tests/programs/aoc_2024/day_12.vi
+++ b/tests/programs/aoc_2024/day_12.vi
@@ -88,7 +88,7 @@ enum Region {
 
 mod Region {
   pub impl show: Show[Region] {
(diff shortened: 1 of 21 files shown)


nilscrm (commented):
I think I would find it more consistent if the trait function and the implementation have to be both marked as functions or both as methods.

tjjfvi:
Then making a trait function a method would be a breaking change to all implementors, which seems like unnecessary churn.

nilscrm:
Hmm, that's also true.

PR #278, opened 2025-06-09
std: standardize container get and at methods

### LINKED ISSUE #217: std: standardize container methods
Currently, we have:
- `list.get(N32) -> &T`
- `array.get(N32) -> &T`
- `map.get(&K) -> Option[&V]`

However:
- both `List::get` and `Array::get` should return `Option`s, as they current exhibit unpredictable behavior if the index is out of bounds
- `get` is not a great name for something that can also mutate
- often one wants to copy the `T` out rather than getting a reference to it

Thus, these interfaces should be standardized:
- `get` returns an `Option[Value]`
- `at` returns a `Option[&Value]`

`.assume()` can be used on the `Option` when the key is known to be present.

nilscrm:
Standardizes access methods to containers with `.at` and `.get`, closes #217.

### DIFF
--- a/tests/programs/aoc_2024/day_04.vi
+++ b/tests/programs/aoc_2024/day_04.vi
@@ -5,7 +5,7 @@ pub fn main(&io: &IO) {
     lines ++= [chars];
   }
 
-  let matches = diamond(
+  let matches = diamond[Char, Char, N32](
     lines,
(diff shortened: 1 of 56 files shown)


nilscrm:
The stats often look fine but sometimes (aoc day_20) total interactions go up by up to 80%, @tjjfvi can you explain this stark increase?

tjjfvi (approved):

Issue #172, opened 2025-02-04
vine: `fn` trait


tjjfvi:
- replace the current `fn(A, B, C) -> D` type with an `fn F(A, B, C) -> D` trait
- `fn foo[T](x: X) -> Y { ... }` creates:
  - a unique type `fn foo[T]`
  - `const foo[T]: fn foo[T]`
  - `impl foo[T]: fn (fn foo[T])(A, B, C) -> D`
- `fn` expressions create unique, unnameable type, with inherent implementation of the appropriate `fn` trait
  - can add `+`/`?`/`*` suffix to `fn` keyword to additionally implement `Fork`/`Drop` (see #171)
- **TBD**: how, exactly, do `fn` impl items work?

```rs
fn map[T, U, F; F*, fn F(T) -> I](list: List[T], f: F) -> List[U] { ... }

struct MulBy(N32);

mod MulBy {
  pub impl call: fn MulBy(N32) -> N32 {
    fn call(MulBy(a): N32, b: N32) -> N32 {
      a * b
    }
  }
}

fn curry_mul(a: N32) -> MulBy {
  MulBy(a)
}

curry_mul(2)(23) // 46
```

vjjft:
In the context of the PR, how is the above example to be viewed(/understood/described/explained)?

PR #282, opened 2025-06-11
vine: rework main handling

### LINKED ISSUE #196: vine: naming conflict with main function in main.vi file
Given the file `main.vi` with
```rust
pub fn main(&io: &IO) { }
```
the following error occurs:

```
error main.vi:1:1 - duplicate definition of `main`
error - cannot find `main` in `::main`
```

This is because the `::main::main` function definition clashes with the internal `::main` node that is used as entry point to the program.

tjjfvi:
The old handling was injecting a `const main: fn(&IO) = ::foo::main` into the root, which is kind of hacky at best, and won't work with #172. Now, the main module/function is just specially tracked/checked.

Also changes the Ivy main to be `::`, from `::main`, and adjusts the set of valid global names to support this.

Resolves #196


vine/src/resolver.rs
```diff
@@ -161,6 +162,61 @@ impl<'core, 'a> Resolver<'core, 'a> {
...
+    let main_mod_name = self.chart.defs[main_mod].name;
+    let main_ident = self.core.ident("main");
+    let path = Path {
+      span,
+      absolute: true,
+      segments: if main_mod_name == main_ident {
+        vec![main_ident]
+      } else {
+        vec![main_mod_name, main_ident]
+      },
```
nilscrm, line 195: Can this not always be `::main_mod::main`, why make the special case when both have the same name?

tjjfvi: Vine auto-collapses things like
```rs
mod foo {
  fn foo() {}
}
```
to define just `foo`, instead of `foo::foo`. Hence the special handling here; if the main module is `main.vi`, its `fn main` will get collapsed to just be `::main`. This is what originally caused #196

ivy/src/lexer.rs
```diff
@@ -38,7 +38,7 @@ pub enum Token {
   N32,
   #[regex(r"[+-][\d\w\.\+\-]+")]
   F32,
-  #[regex(r"(::\p{ID_Continue}+)+")]
+  #[regex(r":[:\p{ID_Continue}]+")]
```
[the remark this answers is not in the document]
tjjfvi: I think I prefer the broader regex. I'm fine reserving anything that starts with a colon as a global name, and it would be nice to have a separator other than `::` that the compiler could use when creating net names, so allowing individual `:` is quite useful.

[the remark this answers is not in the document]
nilscrm: If we can actually make use of the broader syntax, I'm onboard with it.

vine/src/resolver.rs
nilscrm: This could potentially be solved consistently by only merging module names and not function/const names

tjjfvi: This has been part of Vine's module system for a while now, and I think it is quite nice. The standard library uses it extensively; `struct Array` is declared inside `Array.vi`, so without the collapsing, you would need to `use std::data::Array::Array;`. That's pretty common in Rust, actually – e.g. `std::vec::Vec`.

With regard to your example – IMO that seems like a perfectly fine thing to reject. Why would someone want to have both a function `foo` and `foo::foo`?

nilscrm: > This has been part of Vine's module system for a while now, and I think it is quite nice. The standard library uses it extensively; struct Array is declared inside Array.vi, so without the collapsing, you would need to use std::data::Array::Array;. That's pretty common in Rust, actually – e.g. std::vec::Vec.

Yeah this is quite neat.

> With regard to your example – IMO that seems like a perfectly fine thing to reject. Why would someone want to have both a function foo and foo::foo?

Fair, probably pretty rare but unintuitive if someone stumbles upon this.

> (That example would be an error in Rust anyway since you can't have a function and a module with the same name.)

Interesting

PR #286, opened 2025-06-12
vine: support multiple declarations of local in analyzer

tjjfvi:
Previously, the analyzer assumed that each used local was declared exactly once, but this assumption was violated when an identifier pattern matched multiple variants; the local would be declared in each resulting arm.

Now the analyzer supports a local being declared in multiple stages.

I may have had a bit of fun with the test case.


nilscrm (approved):
nice test ;)

PR #294, opened 2025-06-17
vine: require exhaustive pattern matching 

### LINKED ISSUE #134: vine: exhaustiveness checking

tjjfvi:
Resolves #134 

The error message is not very helpful at the moment, but this can be improved in the future. Better than just silently evaluating to an eraser node, at any rate.

### DIFF
--- a/tests/programs/repl/misc.vi
+++ b/tests/programs/repl/misc.vi
@@ -85,3 +85,4 @@ x.drop(); let x;
 F32::inf.abs()
 F32::neg_inf.abs()
 F32::nan.abs()
+do { match Some(1) { Some(_) {} } }
--- a/vine/src/components/resolver.rs
+++ b/vine/src/components/resolver.rs
@@ -336,18 +336,14 @@ impl<'core, 'a> Resolver<'core, 'a> {
   fn resolve_stmts_type(&mut self, span: Span, stmts: &[Stmt<'core>], ty: Type) -> TirExpr {
     let [stmt, rest @ ..] = stmts else {
       let nil = self.types.nil();
-      return if self.types.unify(ty, nil).is_failure() {
-        TirExpr {
-          span,
-          ty,
-          form: Form::Value,
-          kind: Box::new(TirExprKind::Error(
-            self.core.report(Diag::MissingBlockExpr { span, ty: self.types.show(self.chart, ty) }),
-          )),
-        }
+      let kind = if self.types.unify(ty, nil).is_failure() {
+        TirExprKind::Error(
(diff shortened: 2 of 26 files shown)


vine/src/components/resolver.rs
```diff
@@ -334,26 +334,22 @@ impl<'core, 'a> Resolver<'core, 'a> {
...
-      ty,
-      form: Form::Value,
-      kind: Box::new(self._resolve_stmts_type(span, stmts, ty)),
-    }
-  }
-
-  fn _resolve_stmts_type(&mut self, span: Span, stmts: &[Stmt<'core>], ty: Type) -> TirExprKind {
     let [stmt, rest @ ..] = stmts else {
       let nil = self.types.nil();
       return if self.types.unify(ty, nil).is_failure() {
```
nilscrm, line 339:
```suggestion
      let kind = if self.types.unify(ty, nil).is_failure() {
        Box::new(TirExprKind::Error(
          self.core.report(Diag::MissingBlockExpr { span, ty: self.types.show(self.chart, ty) }),
        ))
      } else {
        Box::new(TirExprKind::Composite(Vec::new()))
      };
      return TirExpr { span, ty, form: Form::Value, kind };
```

vine/std/logical/Option.vi
```diff
@@ -56,6 +56,7 @@ pub mod Option {
   pub fn .unwrap[T](self: Option[T]) -> T {
     match self {
       Some(val) { val }
+      None { unsafe::eraser }
```
nilscrm, line 59: For consistency with `Result::unwrap`
```suggestion
      None { unsafe::unreachable() }
```

tests/programs/repl/misc.vi
```diff
@@ -85,3 +85,4 @@ x.drop(); let x;
 F32::inf.abs()
 F32::neg_inf.abs()
 F32::nan.abs()
+do { match Some(1) { Some(_) {} } }
```
nilscrm, line 88: We should add more test for the pattern matching especially for the other patter types.
For example I think currently the exhaustiveness check doesn't work properly for structs
```rust
struct Foo({ foo: Option[N32] });

let foo = Foo({ foo: Some(1) });

match foo {
  Foo({ foo: None }) { 0 }
}
 ```

nilscrm (changes requested):

tests/programs/repl/misc.vi
tjjfvi: That's properly rejected on my machine. Though pasting that into the repl does run into an issue with the repl rollback of variables.

nilscrm (approved):

PR #302, opened 2025-06-28
vine: fork/drop

### LINKED ISSUE #150: vine: repl mishandles variable shadowing
```
io = #io
> let x = 1

io = #io
x = 1
> let x = x + 1

io = #io
> 
```

### LINKED ISSUE #171: vine: `Fork`/`Drop`
```rs
trait Fork[T] {
  fn fork(&self: &T) -> T;
}

trait Drop[T] {
  fn drop(self: T);
}
```

- introduce builtin `Fork`/`Drop` traits
- calls to `Fork` are inserted when a variable is used multiple times
- calls to `Drop` are inserted when a variable is unused
- `Fork` impls can be marked `#[copy]`
  - if an impl of `Fork` and all of its `Fork` impl parameters are marked `#[copy]`, it will be replaced with just a copy
- `Drop` impls can be marked `#[erase]`
  - if an impl of `Drop` and all of its `Drop` impl parameters are marked `#[erase]`, it will be replaced with just an erase
- `T+` shorthand for `Fork[T]`; `T?` shorthand for `Drop[T]`; `T*` shorthand for `Fork[T] & Drop[T]`
- new coercion rules:
  - when a place of type `T` is coerced to a value:
    - if `Fork[T]` is implemented, the place's value will be forked into the place's space
    - if `Drop[~T]` is implemented, the place's space will be dropped
    - otherwise, this is an error
  - when a place of type `T` is coerced to a space:
    - if `Fork[~T]` is implemented, the place's space will be forked into the place's value
    - if `Drop[T]` is implemented, the place's value will be dropped
    - otherwise, this is an error
  - these rules fix the discrepancies between `coerce(~x)` and `~coerce(x)`!
  - **TBD**: error when both rules apply?

tjjfvi:
The forks are all sharpened, the drops are all steepened, and the yaks are all shaved. It's finally here: fork/drop!

- resolves #171 

This is a quite sizable PR. It might be easier to review commit-by-commit, although there are a lot of commits.

Some miscellaneous things that got improved in the course of this PR:
- The pattern matching code has been rewritten, as it was a buggy mess. It now uses types to drive the pattern matching. (That pattern matching code was still from before there were even types.)
- The repl has been significantly reworked. It now handles variables much more sanely (and in a fork/drop aware manner), and some basic commands have been added.
  - resolves #150 
- Implementations can be marked as `#[manual]` to prevent them from being automatically used.

The errors emitted aren't great, but they're *much* better than no errors at all :)

Currently, there's no way to handle variables with multiple implementations of fork/drop; #303.


vine/std/data/List.vi
```diff
@@ -268,7 +262,7 @@ pub mod List {
 
     pub impl drop[T?]: Drop[IntoIter[T]] {
       fn drop(IntoIter[T](len, buf)) {
-        List(len, buf, unsafe::eraser).drop()
+        _ = List(len, buf, unsafe::eraser);
       }
```
nilscrm:
```suggestion
      fn drop(IntoIter[T](_, _)) {}
```
After implementing `Drop` for `Buf[T?]`

vine/src/structures/ast.rs
```diff
@@ -149,6 +149,8 @@ pub enum AttrKind {
   Builtin(Builtin),
   Main,
   Manual,
+  Duplicate,
+  Erase,
```
nilscrm, line 153: What exactly is the purpose of the `duplicate` and `erase` annotation, are they just used to swap the implementations of `fork` and `drop` with the builtins `duplicate` and `erase`?

vine/std/data/List.vi
tjjfvi: `Buf[T]` can't be dropped, since its length is unknown.

vine/src/structures/ast.rs
tjjfvi: Yeah; particularly, when specializing implementations:
- if a `Fork` implementation is marked `#[duplicate]`, and all of its `Fork` impl arguments are `unsafe::duplicate`, then the implementation will be replaced with `unsafe::duplicate`
- if a `Drop` implementation is marked `#[erased]`, and all of its `Drop` impl arguments are `unsafe::erase`, then the implementation will be replaced with `unsafe::erase`

This allows things like `List::fork`, which usually needs to loop over the list, to just duplicate for compatible element types.

This is a pretty significant optimization – see https://github.com/VineLang/vine/pull/302/commits/37551fa3959dcc025feafc7f1c4b7c2dbc378dac for the relevant stats diff.

vine/std/data/List.vi
nilscrm: I'm not sure I understand, what's wrong with
```rust
pub impl drop[T?]: Drop[Buf[T]] {
  fn drop(Buf[T](_, _)) {}
}
```
Wouldn't this just act as a sink that swallows anything you put at the end of the buffer or until it interacts with e.g. the eraser which is put at the end of the List by `into_iter`.

vine/src/components/finder.rs
```diff
@@ -104,6 +111,86 @@ impl<'core, 'a> Finder<'core, 'a> {
...
+      }
+    });
+    error?;
+
+    let pos_flex = pos_fork.is_some() || pos_drop.is_some();
+    let neg_flex = neg_fork.is_some() || neg_drop.is_some();
+
+    if pos_flex && neg_flex {
+      Err(Diag::BiFlexible { span, ty: types.show(self.chart, ty) })?
+    }
```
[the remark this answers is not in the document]
tjjfvi: The new coersion rules are (quoting from #171):
> - when a place of type `T` is coerced to a value:
>   - if `Fork[T]` is implemented, the place's value will be forked into the place's space
>   - if `Drop[~T]` is implemented, the place's space will be dropped
>   - otherwise, this is an error
> - when a place of type `T` is coerced to a space:
>   - if `Fork[~T]` is implemented, the place's space will be forked into the place's value
>   - if `Drop[T]` is implemented, the place's value will be dropped
>   - otherwise, this is an error

These rules mean that (using imaginary syntax) `coerce (~x)` and `~ (coerce x)` are always equivalent, which fixes a lot of issues with using the inverse operator. But they also mean that the coercion is ambiguous if `Fork[T]` and `Drop[~T]` both exist.

vine/std/data/List.vi
nilscrm: Hmm that's sad

vine/src/components/analyzer/effect.rs
```diff
@@ -0,0 +1,155 @@
...
+  PB,
+  RP,
+  RB,
+  RPB,
+  BW,
+  PBW,
+  RBW,
+  RPBW,
+  X,
+}
```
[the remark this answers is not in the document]
tjjfvi: Done. Looks like that's what the `usage.rs` file did before as well.

PR #304, opened 2025-06-28
tooling: stop using workspace lints

HU90m:
This commit doesn't remove any lints from any packages.

This is a slightly cheeky PR. Because I don't really have time to work out how to fix https://github.com/NixOS/nixpkgs/issues/406228 , removing the workspace lints would mean I don't have to maintain a fork for dependent packages. Hopefully this issue will be fixed upstream soon, either by me or someone else.

We are currently running into this issue for in a dependent package.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,8 +1,3 @@
 [workspace]
 resolver = "2"
 members = ["cli", "ivm", "ivy", "lsp", "tests", "util", "vine"]
-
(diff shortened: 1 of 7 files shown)


tjjfvi (approved):

tjjfvi:
(Even if there is not a need for this change, it is probably cleaner IMO to have the lint configuration be package-specific.)

PR #306, opened 2025-06-28
vine: rework iterator api; add `for` loop

### LINKED ISSUE #215: vine: `for` loop
Like rust, syntactic sugar for using the `Iterator` trait:

```rs
for $pat in $iter {
  $body
}
// equivalent to
let it = $iter;
while Iterator::next(&it) is Some($pat) {
  $body
}
```

### LINKED ISSUE #299: vine: change `Iterator` to support non-droppable iterators
Currently, `Iterator` is defined as:
```rs
pub trait Iterator[T, Item] {
  fn .next(self: &T) -> Option[Item];
}
```
However, this presents two issues:
- an iterator could return `None` and then start returning `Some` again
- anything that consumes an iterator must eventually drop it

Since there is no distinction between dropping an iterator early (before it has returned `None`), and dropping it after it is exhausted, it is impossible to have an iterator that must be iterated to completion.

`Iterator` could instead be defined as:
```rs
pub trait Iterator[T, Item] {
  fn .next(self: T) -> Option[(Item, T)];
}
```
This would mean that once an iterator returned `None`, it is definitely exhausted, and furthermore removes the caller's obligation to then drop the iterator. Then a drop bound for the iterator is only necessary if the iteration may be stopped early.

This would make the interface less natural to use, though that would mostly be handled by #215. It might be worthwhile to have a utility 
```rs
fn .frobnicate[T, Item; Iterator[T, Item]](&iter: &Option[T]) -> Option[Item] {
  match iter.next() {
    Some((value, it)) {
      iter = Some(it);
      Some(value)
    }
    None {
      iter = None;
      None
   }
  }
}
```

TBD: is `next` still the best name for this new signature? (Probably not; perhaps something like `Step` would be desired.)

TBD: what should this `frobnicate` utility be called?

tjjfvi:
- resolves #299 
- resolves #215 

Drafted due to dependency on #302


nilscrm (commented):
Nice! I think there are also still quite a few
```rust
while list.pop_front() is Some(el) {}
```
loops that could be converted to
```rust
for el in list.into_iter() {}
```

nilscrm (approved):

PR #309, opened 2025-07-01
vine: transpose

tjjfvi:
"Transposes" the compiler, changing the organization from 'component -> feature' to 'feature -> component', so that there are now distinct files for each high-level language feature, which each implement the corresponding logic across the parser, formatter, charter, resolver, distiller, and/or emitter.


nilscrm (commented):
Code wise, I don't have a super strong preference for one or the other structure.

One thing that we might want to consider though is that with this new structure it will not really be possible to package things separately. I think this would be especially useful for the parser which other tools could rely on.

tjjfvi:
If we need to do that, it could always be done with feature flags.

This new structure feels a lot better to me and I think will be much easier to maintain going forward.

nilscrm (approved):

PR #312, opened 2025-07-01
docs: fix double slash

marvinborner:
Previously the link caused the `types/standard` page not to load CSS correctly. I grepped the source but couldn't find similarly broken paths.


tjjfvi (approved):
thanks!

PR #313, opened 2025-07-03
tests: add F32 cube root test

nilscrm:
Was writing an F32 test case for another project and thought Vine could use it since there are currently only repl tests with F32s.


tjjfvi (approved):
The mandelbrot examples use F32s (and are actually why I added them in the first place), but it can't hurt to have some more :)

PR #314, opened 2025-07-04
vine: overhaul control flow

### LINKED ISSUE #211: vine: change semantics of `loop`
Currently, `loop { ... }` is equivalent to `while true { ... }`. This is undesirable for two reasons:
- by default, it loops indefinitely, but this is not allowed in Vine, so a `break` is required
- in my experience, most uses of `loop` want breaking to be the default, and to only continue in certain circumstances

Thus, change `loop` to not continue the loop by default. `continue` must be explicitly used to continue the loop when desired.

`loop { ... }` would now be equivalent to `while true { ...; break }`.

Really, though, `loop` is the base construct, so it would be more accurate to say `while ... { ... }` is equivalent to `loop { if ... { ...; continue } }`

### LINKED ISSUE #226: vine: `when` expression
Taking inspiration from Kotlin, introduce a `when` expression of the form
```rs
when {
  foo { ... }
  bar { ... }
  baz { ... }
  ...
  else { ... }
}
```

This would be equivalent to
```rs
if foo {
  ...
} else if bar {
  ...
} else if baz {
  ...
} else {
 ...
}
```

`when`s should also be labelable, and should support `break` and `continue` like `match` (#121).

### LINKED ISSUE #259: vine: make `return`/`break`/`continue` statements
Currently, `return`, `break`, and `continue` are expressions. This is inherited from Rust, where this is mostly used in match arms. Because match arms are always blocks in Vine, I don't think there is anywhere where these being expressions is beneficial. It also makes certain patterns more awkward; for example, the following is a type error:
```rs
let foo = match result {
  Ok(value) { value }
  Err(err) { return Err(err); }
};
```
because of the `;` after the `return` expression, the block is typed as evaluating to `()`, rather than generically any `T` like `return` does. Making `return` (and its kin) a statement will resolve this.

### LINKED ISSUE #260: vine: non-divergent let/else
Currently, the `else` block of a `let`/`else` statement must diverge – this constraint is inherited from Rust. Thus, if you want to bail out of a block with a value, one must do something like
```rs
do.foo {
  ...
  let ... = ... else {
    break.foo 123
  };
  ...
}
```

What if, instead, the `else` block did not need to be divergent, and the result of the else block would be the result of the surrounding block? Then the above example could be
```rs
do {
  ...
  let ... = ... else {
    123
  };
  ...
}
```
I think this reads well/better.

### LINKED ISSUE #266: vine: `while`/`else`
(This issue assumes the semantics of `loop` from #211.)

Currently, `while $cond { $body }` is equivalent to:
```rs
loop {
  if $cond {
    $body
    continue
  }
}
```

Add support for `while` loops with an `else` block, written `while $cond { $body } else { $else }`, equivalent to:
```rs
loop {
  if $cond {
    $body
    continue
  } else {
    $else
  }
}
```

That is, the `else` block of a `while` loop executes if the condition of the while fails (but not when `break` is used in the loop).

A `while`/`else` loop evaluates to the value used to break the loop, if the loop is broken, or otherwise the value of the `else` block.

This can be useful for searching; for example:
```rs
let magic_number = do {
  let i = 0;
  while i < max {
    if is_magic_number(i) {
      break Some(i);
    }
    i += 1;
  } else {
    None
  }
};
```

Once `for` loops are added (#215), a `for`/`else` should exist as well, with similar semantics. For example:
```rs
let magic_number =
  for i in 0..max {
    if is_magic_number(i) {
      break Some(i);
    }
  } else {
    None
  }
```

### LINKED ISSUE #296: vine: allow annotation of return type of any block
e.g.
```rs
do -> N32 {
  123
}

match result -> Option[_] {
  Ok(value) { Some(value) }
  Err(_) { None }
}
```

### LINKED ISSUE #297: vine: `let`/`else match`
An extension to `let`/`else` that allows matching the value that failed to match the `let` pattern.

Currently, we have e.g.
```rs
let Some(value) = option else {
  return "no value found";
};
// rest of the block
```
which is equivalent to
```rs
match option {
  Some(value) {
    // rest of the block
  }
  None {
    return "no value found";
  }
}
```
But when matching on e.g. a `Result`, there is no way to use the `Err` value. With `let`/`else match`, something like
```rs
let Ok(value) = result else match {
  Err(err) {
    return "error: {err}";
  }
};
// rest of the block
```
would be equivalent to:
```rs
match result {
  Ok(value) {
    // rest of the block
  }
  Err(err) {
    return "error: {err}";
  }
}
```

### LINKED ISSUE #298: vine: control flow overhaul
- #211 
- #266 
- #215 
- #226 
- #296 
- #260 
- #297
- #121 
- #259 
- #251

tjjfvi:
- resolves #298 
  - `loop` now no longer has an implicit continue
    - resolves #211
  - `while` and `for` both support `else` blocks, which allow them to evaluate to a value
    - resolves #266
  - optimization: variables with type `()` or `{}` are now no longer passed between stages
    - the `while`/`else` and `for`/`else` change added a number of `()`-type variables, so this optimization was necessary to not regress performance for these
  - `break` and `continue` support specifying a keyword to break, instead of a label (e.g. `break.do`), making a lot of labels unnecessary
  - added `when` expressions
    - resolves #226
    - removed the `else if` construct; one can instead use a `when` or add braces around the `if`
  - all control flow constructs that can evaluate to a value now support annotating their return type
    - resolves #296
  - `let`/`else` no longer requires that the else block explicitly diverges
    - resolves #260
  - added `let`/`else match`
    - resolves #297
  - made `return`/`break`/`continue` statements
    - resolves #259

Likely easiest to review commit-by-commit, as there are a lot of changes here.


tests/programs/aoc_2024/day_02.vi
```diff
@@ -9,11 +9,15 @@ pub fn main(&io: &IO) {
...
-      dampened_safe_count += 1;
+    when {
+      is_safe(report) {
+        safe_count += 1;
+        dampened_safe_count += 1;
+      }
+      problem_dampener(report) {
+        dampened_safe_count += 1;
+      }
+      _ {}
```
nilscrm, line 20: In this instance I prefer the `if else` version as it doesn't need the extra empty arm.

tests/programs/aoc_2024/day_14.vi
nilscrm, line 1: Here I think the `if else` version is nicer as well.

vine/examples/guessing_game.vi
```diff
@@ -24,18 +24,22 @@ pub fn main(&io: &IO) {
...
+        io.println("Too high.");
+        continue;
+      }
+      _ {
+        io.println("You got it " ++ if guesses == 1 {
+          "immediately?!"
+        } else {
+          "in {guesses} guesses!"
+        });
+      }
```
nilscrm, line 42: This is when I really like having `when` :)

tests/programs/repl/misc.vi
```diff
@@ -91,3 +91,5 @@ let x = 1
 let x = x + 1
 /clear x
 match (Some(1), 2) { (None, _) { None } x { Some(x) } }
+let Ok(n) = Ok(46) else match { Err(e) { e } }; "{n}"
+let Ok(n) = Ok("abc") else match { Err(e) { e } }; "{n}"
```
nilscrm, line 95: Can we get one more test where the `else match` block is executed.

nilscrm (commented):
Very nice, lots of good stuff!

Here're my thoughts on the different features:
`else` for `while` and `for`: Very handy, I like it

`when`: really useful, love it

removal of `else if`: I don't really like it. Event with the `when` statement there are still situation where `else if` is nicer I think (see comments). Does keeping it come at any cost?

`let / else match`: I like it especially for thinks like errors

no implicit continue in loops: I strongly dislike it. I think the semantics get less intuitive and this is especially bad in `loop`s that are basically `while` loops but break in the the middle or at the end, e.g. `find_primes.vi`. One could make the argument that with a mandatory continue at least the compiler would warn you, but I think that would add verbosity. Seeing all the changes there are also already more `continue`s added than `break`s removed.

keywords as break target: Quite useful, haven't seen that before but I like it

`return, break, continue`: As discussed before, solid choice

annotation of return type of control flow structures: Probably very useful in cases where type inference is lost

non-divergence `let /else`: I'm unsure, for me it's a little too implicit and it sounds somewhat like you're choosing an alternative value for the let assignment.

tests/programs/aoc_2024/day_02.vi
```diff
@@ -32,26 +32,28 @@ fn is_safe(report: List[N32]) -> Bool {
         level - last
       };
       if !(1 <= diff <= 3) {
-        return false;
+        break false;
       }
       if order is Some(order) && order != (last > level) {
-        return false;
+        break false;
```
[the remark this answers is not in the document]
tjjfvi: I've changed the `return`s to `break`s where it matches the underlying semantics better. Here, for example, the loop is performing a search, and so when we encounter an unsafe value, we halt the search by breaking the loop. The fact that it is the return value of the function is not relevant to the search – one could imagine, for example, refactoring this such that it returns both the boolean from the search and some other data, at which point one would want to store the value from the loop in a variable, rather than having an early return. 

A good example of where this is useful is in `day_06.vi` – before, there was an early `return (false, distinct), and a regular return of `(true, distinct)`. But it's more semantically correct to have the loop just evaluate to a boolean, and have the return of `(looped, distinct)` in one place.

tjjfvi:
Regarding `else if`: in both of the examples you mentioned preferring `else if`, it seemed like the main complaint was the empty `_ {}` arm. I think it would make sense to stop requiring this, as it adds no additional information. Would that resolve your concerns?

tests/programs/repl/misc.vi
tjjfvi: Oops, that last one was supposed to be `Err("abc")`. Good catch :)

nilscrm:
> Regarding `else if`: in both of the examples you mentioned preferring `else if`, it seemed like the main complaint was the empty `_ {}` arm. I think it would make sense to stop requiring this, as it adds no additional information. Would that resolve your concerns?

Yeah I think it makes sense to not require an empty arm and only have `match` statements check for exhaustiveness. I think it would also remove most of my concerns, but on the other hand I feel like we're not giving anything up by supporting the very standard `else if`. One situation I can see it being useful is when you have a few nested conditions where `when`s would introduce twice the level of indentation.

tjjfvi:
> Yeah I think it makes sense to not require an empty arm and only have `match` statements check for exhaustiveness.

👍

> I feel like we're not giving anything up by supporting the very standard `else if`.

I don't think it's good to have two different constructs that do the same thing; it creates questions of when which should be used, potentially requires converting between them, and makes code written in the language less consistent.

> One situation I can see it being useful is when you have a few nested conditions where `when`s would introduce twice the level of indentation.

If one has nested conditions to the point that the increased indentation is excessive, a refactor is probably warranted.

tjjfvi:
I've removed the requirement for the terminal arm in `when`s. After discussing some more offline, I'd like to move forward with the changes to `loop` and `let`/`else`.

Issue #273, opened 2025-06-08
vine: stack overflow on recursive type alias


tjjfvi:
This used to be checked; not sure where this went...

PR #316, opened 2025-07-05
fix: #273 stack overflow

falconnor4:
The problem was that the `resolve_type_alias` function would immediately start resolving the alias type without first marking that it was currently being resolved, when it encountered a recursive reference it would infinitely recurse instead of detecting the cycle. The fix was to mark the alias as "being resolved" (with `None`) before starting resolution, so recursive references can be detected early.

### DIFF
--- a/vine/src/features/type_.rs
+++ b/vine/src/features/type_.rs
@@ -68,7 +68,7 @@ impl<'core> Charter<'core, '_> {
 impl<'core> Resolver<'core, '_> {
   pub(crate) fn resolve_type_alias(&mut self, alias_id: TypeAliasId) {
     match self.sigs.type_aliases.get(alias_id) {
-      Some(TypeAliasState::Resolved(_)) => return,
+      Some(TypeAliasState::Resolved(_)) => {}
       Some(TypeAliasState::Resolving) => {
         // Cycle detected
         let alias_def = &self.chart.type_aliases[alias_id];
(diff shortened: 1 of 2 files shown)


tjjfvi:
Thanks for taking a look at this!

This will currently incorrectly reject code such as:
```rs
type A = C;
type B = N32;
type C = B;
```
These will get assigned type ids in order. Resolving `A` will then need to resolve `C`, which will fill the vector with `None`s up to `C`, thereby setting `B` to `None`. When `C` then needs to resolve `B`, it will think there has been a circular type alias.

Instead of just having an option, there should probably be an enum with variants `Unresolved`, `Resolving`, and `Resolved`, so that only `Resolving` is reported as a cycle.

tests/programs/fail/recursive_type_alias.vi
```diff
@@ -0,0 +1,13 @@
+// Test case for recursive type alias detection
+type A = A;
+
+// Test case for indirect recursion
+type B = C;
+type C = B;
+
+// Test case for self-referential with complex structure  
+type F = Array[F];
```
tjjfvi, line 9: This is currently creating an error because `Array` was not imported; I would suggest using `List` instead.

falconnor4: Thanks! I'll get to that when I get to my computer. I completely overlooked that.

vine/src/features/type_.rs
```diff
@@ -67,14 +67,32 @@ impl<'core> Charter<'core, '_> {
...
 impl<'core> Resolver<'core, '_> {
   pub(crate) fn resolve_type_alias(&mut self, alias_id: TypeAliasId) {
-    if let Some(Some(_)) = self.sigs.type_aliases.get(alias_id) {
-      return;
+    match self.sigs.type_aliases.get(alias_id) {
+      Some(TypeAliasState::Resolved(_)) => return,
+      Some(TypeAliasState::Resolving) => {
+        // Cycle detected
+        let alias_def = &self.chart.type_aliases[alias_id];
+        self.core.report(Diag::RecursiveTypeAlias { span: alias_def.ty.span });
```
tjjfvi, line 75: `report` is called later in `let error_type` so this currently duplicates the error message
```suggestion
```

vine/src/features/type_.rs
```diff
@@ -67,14 +67,32 @@ impl<'core> Charter<'core, '_> {
...
+        let slot = self.sigs.type_aliases.get_or_extend(alias_id);
+        let error_type =
+          self.types.error(self.core.report(Diag::RecursiveTypeAlias { span: alias_def.ty.span }));
+        *slot = TypeAliasState::Resolved(
+          self.types.export(|t| TypeAliasSig { ty: t.transfer(&error_type) }),
+        );
+        return;
+      }
+      _ => {}
```
tjjfvi, line 84: I think it would be cleaner to move the remaining logic of the function into this arm and remove the two early `return`s.

tjjfvi (commented):
looking good, just a few nits

Issue #169, opened 2025-02-04
vine: impl derivation


tjjfvi:
Introduce `Tuple` and `Object` traits
```rs
trait Tuple[Tuple, Init, Rest] {
  fn deconstruct(tuple: Tuple) -> (Init, Rest);
  fn reconstruct(init: Init, rest: Rest) -> Tuple;
}

trait Object[Object, Init, Rest] {
  const key: String;
  fn deconstruct(object: Object) -> (Init, Rest);
  fn reconstruct(init: Init, rest: Rest) -> Object;
}
```
The compiler would automatically implement `Tuple` and `Object` for all non-empty tuple and object types; e.g. `Tuple[(A, B, C), A, (B, C)]` and `Object[{ a: A, b: B, c: C }, A, { b: B, c: C }]`. The order in which object types are broken down would not be guaranteed.

This allows traits to be inductively implemented for all tuple/object types by providing impls for `()`, `{}`, and types implementing `Tuple`/`Object`.

For struct types, introduce an additional trait
```rs
trait Struct[Struct, Data] {
  const name: String;
  fn deconstruct(self: Struct) -> Data;
  fn reconstruct(data: Data) -> Struct;
}
```
Not sure how best to make implementations that use this impl to be opt-in.

Also not sure how best to handle enums.

PR #318, opened 2025-07-07
vine: tuple/object impl derivation

tjjfvi:
Implements the `Tuple` and `Object` traits from #169, allowing traits to be implemented for all tuple and object types, as well as a `Composite` trait that abstracts over `Tuple` and `Object` for things like `Fork` and `Drop` which do not need to differentiate between tuples and objects. These traits are now used for implementing `Fork`, `Drop`, `Show`, `Eq`, and `Ord`.


nilscrm (approved):
This is so nice and missing from so many languages

PR #322, opened 2025-07-10
vine: indirect impls

### LINKED ISSUE #174: vine: impl aliases

tjjfvi:
Allows traits to be implemented indirectly, by specifying another implementation to delegate to. E.g.
```rs
pub impl fork: Fork[N32] = unsafe::duplicate;
```

This is not an alias, as the indirect implementation and the implementation it delegates to are considered distinct.

This is thus mostly useful to use with `#[manual]` implementations, as otherwise this will usually result in multiple implementation errors.

- resolves #174

### DIFF
--- a/vine/src/structures/checkpoint.rs
+++ b/vine/src/structures/checkpoint.rs
@@ -277,7 +277,6 @@ impl<'core> Resolutions<'core> {
     consts.truncate(checkpoint.concrete_consts.0);
     fns.truncate(checkpoint.concrete_fns.0);
     impls.truncate(checkpoint.impls.0);
-    impls.truncate(checkpoint.impls.0);
     revert_idx(main, checkpoint.fragments);
   }
 }


vine/src/structures/checkpoint.rs
```diff
@@ -273,18 +267,17 @@ impl<'core> Signatures<'core> {
...
 impl<'core> Resolutions<'core> {
   fn revert(&mut self, checkpoint: &Checkpoint) {
-    let Resolutions { consts, fns, direct_impls, indirect_impls, main } = self;
+    let Resolutions { consts, fns, impls, main } = self;
     consts.truncate(checkpoint.concrete_consts.0);
     fns.truncate(checkpoint.concrete_fns.0);
-    direct_impls.truncate(checkpoint.direct_impls.0);
-    indirect_impls.truncate(checkpoint.indirect_impls.0);
+    impls.truncate(checkpoint.impls.0);
+    impls.truncate(checkpoint.impls.0);
```
nilscrm, line 280: I think this accidentally stayed duplicated during the refactor.
```suggestion
```

nilscrm (approved):
Looks good

PR #323, opened 2025-07-10
vine: generic modules and trait items

### LINKED ISSUE #209: vine: generics on `mod`
```rs
pub mod foo[T, U; Ord[T]] {
  pub const bar: N32 = 1;
  pub const baz[...]: Map[T, U] = Map::empty[T, U];
  pub fn qux[... V; ToString[T], ToString[V]](/* ... */) { /* ... */ }
}
```
would be equivalent to
```rs
pub mod foo {
  pub const bar: N32 = 1;
  pub const baz[T, U; Ord[T]]: Map[T, U] = Map::empty[T, U];
  pub fn qux[T, U, V; Ord[T], ToString[T], ToString[V]](/* ... */) { /* ... */ }
}
```

tjjfvi:
- allows modules to be given generic parameters, which are inherited by children whose generic parameters start with `...`
  - resolves #209 
- allows items within traits and implementations to be generic; e.g. trait functions can now take `fn` parameters

(The latter feature, which is needed for origins, depends on the same technical foundation as the former; hence why this is bundled together.)


nilscrm (commented):
I think generic trait items are currently not used anywhere. I think it makes sense to add a test case for that.

tjjfvi:
I updated `par.vi` to use generic trait items, as it had previously taken a closure parameter, but that had to be changed in #293.

nilscrm (approved):

PR #332, opened 2025-07-24
std: implement F32::ln

nilscrm:
Implement F32::ln based on the Rust [libm implementation](https://docs.rs/libm/latest/src/libm/math/logf.rs.html).


nilscrm:
Close in favor of #339

PR #333, opened 2025-07-24
ivm: add f32_sqrt extrinsic function

nilscrm:
Since almost all modern hardware support sqrt on floats, I think it makes sense to add it as a native extrinsic function.


tjjfvi:
> Since almost all modern hardware support log on floats, I think it makes sense to add it as a native extrinsic function.

Did you mean to say `sqrt` here?

nilscrm:
> Did you mean to say `sqrt` here?

ah yes

nilscrm:
Close in favor of #339

PR #337, opened 2025-07-26
std: fix I32::to_string for negative values

nilscrm:


tjjfvi (approved):
this is embarrassing...

PR #343, opened 2025-07-30
Implement Smoothsort for Array[T] (closes #185)

### LINKED ISSUE #185: std: sorting
- use `Ord` trait
- not quadratic
- ~~work natively on `Array`s~~

chalcolith:
Smoothsort is a sorting algorithm by Edsger Dijkstra. It is O(n) for mostly-sorted data, and O(n log n) in the worst case. This implementation is based on Keith Schwartz's at https://www.keithschwarz.com/smoothsort/

As this is my first attempt at Vine, please let me know how to best fit this into the standard library, and how to make it more idiomatic.

One thing I noticed was that `loop` doesn't seem to actually loop; its body only ran once.


chalcolith:
I have taken this out of the standard library and made it a standalone test program.  I believe I have addressed all the feedback, except that when I try to use `for _ in 0..n` I get an error "cannot find impl of trait `Iterator[Range[Inclusive[N32], Exclusive[N32]], ?30]`"; do I need to import something to use this?

tjjfvi:
> when I try to use `for _ in 0..n` I get an error "cannot find impl of trait `Iterator[Range[Inclusive[N32], Exclusive[N32]], ?30]`"; do I need to import something to use this?

I think you just need to rebase onto the latest `main`; the PR implementing that was merged quite recently: #342

tests/programs/array_smoothsort.vi
```diff
@@ -0,0 +1,277 @@
...
+        // case 3: add a heap of order 1
+        shift_left(&trees0, &trees1, smallest_tree - 1);
+        trees0 = trees0 | 1;
+        smallest_tree = 1;
+      }
+    }
+  }
+
+  // check if the new tree is in its final position
+  let isLast = false;
```
tjjfvi, line 81: this should be `snake_case`

tests/programs/array_smoothsort.vi
```diff
@@ -0,0 +1,277 @@
...
+
+    // adjust the bitmap
+    loop {
+      shift_right(&trees0, &trees1, 1);
+      smallest_tree += 1;
+
+      if (trees0 & 1) == 1 {
+        break;
+      }
+      continue;
```
tjjfvi, line 158:
```suggestion
      if trees0 & 1 != 1 {
        continue;
      }
```

tjjfvi (commented):
Looking good! Noticed two small things.

I think the only thing that wasn't addressed was [using `let fn`s for the helpers](https://github.com/VineLang/vine/pull/343#discussion_r2243407121). I think this would make the code a bit more readable, and it would be nice to have a more complex test of recursive `let fn`s :)

chalcolith:
OK, I've moved the helper functions inside the main one.

tjjfvi (approved):

tjjfvi:
Looks good; thanks!

PR #347, opened 2025-08-10
std: implement element-wise operations on tuples

nilscrm:
Allows for element-wise arithmetic and logical operations of tuples.

* Rename `vectorized` to `element_wise` to not confuse e.g. the `Mul` implementation with vector dot products.
* Finder will construct Tuple types if the `Rest` of the tuple is known
* add `as _` imports for anonymous imports of implementations

### DIFF
--- a/tests/programs/aoc_2024/day_10.vi
+++ b/tests/programs/aoc_2024/day_10.vi
@@ -56,12 +56,9 @@ pub fn main(&io: &IO) {
   io.println("Ratings: {ratings}");
 }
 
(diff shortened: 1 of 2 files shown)


tjjfvi (commented):
Two high-level thoughts:
- I think it should be `elementwise` rather than `element_wise`
- I only had individual imports of `add`, `sub`, etc. before because there was no way to be able to import them all together. Now that that is here, perhaps it would be nicest to just `use std::ops::elementwise;`, and get all of the implementations?

PR #350, opened 2025-08-14
test: conway's game of life

tjjfvi:
Uses a much cleaner variant of the network-communication technique used in some of the AoC programs. Could be worth extracting into a library at some point.


tests/programs/life.vi
```diff
@@ -0,0 +1,366 @@
...
+    unsafe::erase(recv);
+  }
+}
+
+pub struct Pipe[T]((Send[T], Recv[T]));
+pub mod Pipe {
+  pub const new[T]: Pipe[T] = do {
+    let stream;
+    Pipe(Send(~stream), Recv(stream))
+  };
```
nilscrm, line 244: Is there a reason to use a const instead of a `new` function. For readability I think a `new` function is clearer.

nilscrm (approved):

PR #365, opened 2025-10-25
lsp: symbols, definitions, references, and hover

### LINKED ISSUE #99: vine: lsp hover

### LINKED ISSUE #100: vine: lsp go-to-def

tjjfvi:
Implements:
- workspace symbols
- document symbols
- go to definition (resolves #100)
- go to references
- document highlight (highlighting all occurences of a symbol in the document)
- hover (resolves #99)
- doc comment syntax

Additionally, the lsp will no longer recompile the standard library on every save, which saves about 500ms on my machine.

Also adds a basic textmate grammar for Vine to the vscode extension, mostly for the sake of highlighting in hover. It only highlights things lexically (not syntactically), but it does correctly handle nested comments and string interpolation:
image

### DIFF
--- a/vine/src/compiler.rs
+++ b/vine/src/compiler.rs
@@ -59,13 +59,9 @@ impl Compiler {
     }
   }
 
-  pub fn compile(&mut self, revert: bool, hooks: impl Hooks) -> Result<Nets, Vec<Diag>> {
+  pub fn compile(&mut self, hooks: impl Hooks) -> Result<Nets, Vec<Diag>> {
     let checkpoint = self.checkpoint();
-    self._compile(hooks, &checkpoint).inspect_err(|_| {
-      if revert {
-        self.revert(&checkpoint);
-      }
-    })
+    self._compile(hooks, &checkpoint)
   }
 
   fn _compile(
--- a/cli/src/vine_cli.rs
+++ b/cli/src/vine_cli.rs
@@ -76,7 +76,7 @@ impl CompileArgs {
       compiler.loader.load_mod(&lib, &mut compiler.diags);
     }
 
-    match compiler.compile(false, ()) {
+    match compiler.compile(()) {
       Ok(nets) => nets,
       Err(diags) => {
         eprintln!("{}", compiler.loader.print_diags(&diags));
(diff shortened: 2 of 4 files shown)


tjjfvi:
(Based on #410 & thus draft, but ready to review.)

vine/src/components/charter.rs
```diff
@@ -142,7 +142,7 @@ impl Charter<'_> {
     if let Some(def) = def {
       for subitem in subitems {
         if !matches!(subitem.vis, Vis::Private) {
-          self.diags.report(Diag::VisibleSubitem { span: subitem.span });
+          self.diags.report(Diag::VisibleSubitem { span: subitem.name_span });
```
nilscrm: I think this error looks nicer with `span` than `name_span` since then the `pub` is actually highlighted as well.

vine/src/features/trait_.rs
```diff
@@ -63,7 +64,7 @@ impl Charter<'_> {
       if !self.enabled(&subitem.attrs) {
         continue;
       }
-      let span = subitem.span;
+      let span = subitem.name_span;
```
nilscrm: I think this is nicer with `span` as well.

nilscrm (commented):
This is so nice :)

vine/src/components/charter.rs
tjjfvi: I don't think it's good to have errors that cover the entire item (e.g. the entire function body), so I changed all of the errors on items to use the span of the name. I think making the error spans more precise (e.g. just highlighting the visibility, etc.) is a larger project.

vine/src/compiler.rs
```diff
@@ -59,10 +59,12 @@ impl Compiler {
     }
   }
 
-  pub fn compile(&mut self, hooks: impl Hooks) -> Result<Nets, Vec<Diag>> {
+  pub fn compile(&mut self, revert: bool, hooks: impl Hooks) -> Result<Nets, Vec<Diag>> {
```
nilscrm: Since the revert logic is only used once in the repl, I would move it there.

PR #371, opened 2025-11-03
lsp: support formatting

nilscrm:
This way the lsp advertises the vine formatter and it doesn't need to be set up separately with dprint but works out of the box.

### DIFF
--- a/lsp/src/lib.rs
+++ b/lsp/src/lib.rs
@@ -144,17 +144,16 @@ impl LanguageServer for Backend {
     let arenas = CoreArenas::default();
     let core = Core::new(&arenas, false);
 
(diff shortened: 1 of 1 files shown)


lsp/src/lib.rs
```diff
@@ -108,6 +111,52 @@ impl LanguageServer for Backend {
...
+      match docs.get(&uri) {
+        Some(text) => text.clone(),
+        None => return Ok(None),
+      }
+    };
+
+    let arenas = CoreArenas::default();
+    let core = Core::new(&arenas, false);
+
+    if let Ok(formatted) = core.fmt(&src) {
```
tjjfvi: this might be nicer as a `let`/`else`

tjjfvi (approved):

Issue #223, opened 2025-05-12
std: `Nat` & `Int`


tjjfvi:
Arbitrary-precision natural number and integer types, respectively.

I think the best implementation would be to store them as a tuple `(order, tree)`, where `order` is an `N32` and `tree` is a perfect binary tree of `N32`s of depth `order` (thus having `2^order` leaves of type `N32`). Thus a `Nat` of order `n` would have `32 * 2^order` bits. The maximum representable value would then be `2^2^(2^32 + 4) - 1`, which is absurdly large.

`Int` would be represented similarly, using 2's complement. I'm not sure what the best way to represent the sign bit is. The most significant bit of the tree would work but is hard to access. It could also be a bit in the `order`.

tjjfvi:
Ideally the addition algorithms for these types could take inspiration from carry-lookahead adders in order to be parallelizable.

The Karatsuba algorithm would be very convenient for multiplication, given that the digits are already nicely divided into halves.

If any asymptotically better algorithms are more costly at small orders, then the implementation could branch on the order to choose whichever algorithm would be faster. This order threshold could be rather precisely determined thanks to the number of interactions being deterministic.

vjjft:
> This order threshold could be rather precisely determined thanks to the number of interactions being deterministic.

Would this determination be potentially dependent upon runtime implementation and tested hardware?

If so, might there be merit in allowing overriding the thresholds at runtime?

PR #379, opened 2025-11-05
std: implement Nat

nilscrm:
see #223

### DIFF
--- a/tests/programs/repl/Nat.vi
+++ b/tests/programs/repl/Nat.vi
@@ -5,36 +5,36 @@ use std::debug::Show;
 123 as Nat
 
 42 as Nat + 4 as Nat
-"{Nat::parse("1111111111111111111111111111").unwrap() + Nat::parse("55555555555555555555").unwrap()}"
+Nat::parse("1111111111111111111111111111").unwrap() + Nat::parse("55555555555555555555").unwrap()
 Nat::from_parts([N32::maximum, N32::maximum, N32::maximum, N32::maximum]) + Nat::from_parts([N32::maximum, N32::maximum])
-"{Nat::parse("18446744073709551615").unwrap() + Nat::parse("18446744073709551617").unwrap()}"
+Nat::parse("18446744073709551615").unwrap() + Nat::parse("18446744073709551617").unwrap()
 
 100 as Nat - 42 as Nat
 100 as Nat - 0 as Nat
 0 as Nat - 100 as Nat
 Nat::parse("1111111111111111111111111111").unwrap() - Nat::parse("1111111111111111111111111111").unwrap()
-"{Nat::parse("6666666666666666666666666666").unwrap() - Nat::parse("2222222222222").unwrap()}"
+Nat::parse("6666666666666666666666666666").unwrap() - Nat::parse("2222222222222").unwrap()
 
-"{(0 as Nat).div_rem_n32(1).show()}"
-"{(0 as Nat).div_rem_n32(123_456_789).show()}"
-"{(5 as Nat).div_rem_n32(1).show()}"
-"{(5 as Nat).div_rem_n32(2).show()}"
-"{(5 as Nat).div_rem_n32(10).show()}"
-"{(123 as Nat).div_rem_n32(10).show()}"
-"{(123 as Nat).div_rem_n32(124).show()}"
-"{(124 as Nat).div_rem_n32(124).show()}"
-"{(1 as Nat).div_rem_n32(10).show()}"
-"{(1 as Nat).div_rem_n32(4_294_967_295).show()}"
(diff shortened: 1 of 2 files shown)


vine/std/numeric/Nat.vi
```diff
@@ -0,0 +1,301 @@
...
+  pub impl add: Add[Nat, Nat, Nat] {
+    fn add(Nat(a), Nat(b)) -> Nat {
+      let result = [];
+      let carry = 0;
+      let remaining = loop {
+        match (a.pop_front(), b.pop_front()) {
+          (Some(a_num), Some(b_num)) {
+            let lo = a_num + b_num;
+            let sum = lo + carry;
+            carry = N32::add_high(a_num, b_num) + N32::add_high(sum, carry);
```
tjjfvi: This is wrong, should be
```suggestion
            carry = N32::add_high(a_num, b_num) + N32::add_high(lo, carry);
```

vine/std/numeric/Nat.vi
```diff
@@ -0,0 +1,301 @@
...
+            } else {
+              a_num < b_num
+            };
+
+            if !~trim {
+              result.push_back(diff);
+            }
+            if diff != 0 {
+              ~trim = false;
+            }
```
tjjfvi: This is done a few times; I think it would be nice for this to be extracted to a `let fn`

vine/std/numeric/Nat.vi
```diff
@@ -0,0 +1,301 @@
...
+            let borrow = num == 0;
+            continue;
+          }
+          _ {
+            // Underflow
```
tjjfvi: For consistency with above
```suggestion
            // underflow
```

vine/std/numeric/Nat.vi
```diff
@@ -0,0 +1,301 @@
...
+  pub impl mul_n32: Mul[Nat, N32, Nat] {
+    fn mul(Nat(a), n: N32) -> Nat {
+      if n == 0 {
+        return 0 as Nat;
+      }
+      let carry = 0;
+      let result = [];
+      for num in a.into_iter() {
+        let N64(lo, hi) = N64::mul_n32_n32(num, n);
+        result.push_back(lo + carry);
```
tjjfvi: This addition could overflow. Example where this fails:
```
> "{Nat::parse("4294967298").unwrap() * 4294967295}"
"4294967294"
```

vine/std/numeric/Nat.vi
```diff
@@ -0,0 +1,301 @@
...
+  pub impl rem_n32: Rem[Nat, N32, N32] {
+    fn rem(a: Nat, d: N32) -> N32 {
+      a.div_rem_n32(d).1
+    }
+  }
+
+  pub fn .div_rem_n32(a: Nat, d: N32) -> (Nat, N32) {
+    if d == 0 {
+      debug::error("Division by zero");
+      unsafe::eraser
```
tjjfvi:
```suggestion
      return unsafe::eraser;
```

vine/std/numeric/Nat.vi
```diff
@@ -0,0 +1,301 @@
...
+
+  pub fn parse(str: String) -> Option[Nat] {
+    if str.len() == 0 {
+      return None;
+    }
+    let num = 0 as Nat;
+    for digit in str.into_iter() {
+      if digit == '_' {
+        continue;
+      }
```
tjjfvi: I don't think this is done for `N32` and `N64`, but perhaps it should be.

vine/std/numeric/Nat.vi
```diff
@@ -0,0 +1,301 @@
...
+      let (qi, ri) = N64::div_rem_n32(cur, d);
+
+      q.push_front(qi as N32);
+      rem = ri;
+    }
+
+    (Nat(q), rem)
+  }
+
+  pub impl from_n32: Cast[N32, Nat] {
```
tjjfvi: Would also be nice to have `Cast[N64, Nat]`, `Cast[Nat, N32]`, and `Cast[Nat, N64]`.

tjjfvi (changes requested):

vine/std/numeric/Nat.vi
```diff
@@ -0,0 +1,301 @@
...
+    let nums = a!.reversed();
+
+    let Some(leading_num) = nums.pop_front() else {
+      return (Nat::zero, 0);
+    };
+    let q = [];
+    let (N64(qi, _), rem) = N64::div_rem_n32(leading_num as N64, d);
+    if qi != 0 {
+      q.push_front(qi);
+    }
```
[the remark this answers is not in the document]
tjjfvi: Ah, good point. In that case I think that can be simplified to use `/` and `%` on `N32`s directly instead of using `N64::div_rem_n32`

PR #391, opened 2025-11-09
std: optimize/simplify `Nat::{add,sub,mul}`

tjjfvi:


nilscrm (approved):
I think we can completely get rid of `Nat::zero` now that we have arbitrary number literals

PR #396, opened 2025-11-11
std: implement gcd for Nat

nilscrm:


vine/std/numeric/Nat/Nat.vi
```diff
@@ -378,4 +378,41 @@ pub mod Nat {
...
+    }
+    let a_zeros = a.trailing_zeros();
+    a >>= a_zeros;
+    let b_zeros = b.trailing_zeros();
+    b >>= b_zeros;
+    let common_zeros = N32::min(a_zeros, b_zeros);
+
+    loop {
+      if a > b {
+        (a, b) = (b, a);
```
tjjfvi: I'd prefer `swap(&a, &b)` here, but it's not essential

PR #398, opened 2025-11-13
vine: use tree sitter grammar for vs code syntax highlighting

### LINKED ISSUE #235: vine: syntax highlighting
Currently we just steal Rust's syntax highlighting, which is convenient, but is becoming increasingly less accurate.

nilscrm:
Resolves #235 

This uses the semantic tokens API to provide syntax highlighting. This means we don't have to additionally maintain a TextMate grammar or fall back on the rust grammar. Moreover, since tree sitter is an actual parser, we get better highlighting in situations like string interpolation.

### DIFF
--- a/lsp/tree-sitter-vine/grammar.js
+++ b/lsp/tree-sitter-vine/grammar.js
@@ -96,6 +96,38 @@ module.exports = grammar({
     $._impl,
   ],
 
+  reserved: {
+    global: _ => [
+      "pub",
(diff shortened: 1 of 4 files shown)


vine/examples/hello_world.vi
```diff
@@ -1,4 +1,5 @@
 
+
```
nilscrm:
```suggestion
```

PR #400, opened 2025-11-15
lsp: add vine grammar for comments

nilscrm:
This way VS Code is aware of how comments work in Vine and does not do bracket matching in them, which marks them red, e.g. for

```
// x will be in the range [0, 2)
```

### DIFF
--- a/lsp/client/src/ivy-grammar.json
+++ b/lsp/client/src/ivy-grammar.json
@@ -1,8 +1,8 @@
 {
   "name": "ivy",
   "patterns": [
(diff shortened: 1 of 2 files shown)


nilscrm:
> This breaks with strings:
> 
> ```rust
> io.println("Hello//world");
>                       // ^^ highlighted as a comment
> ```

This is definitely worse than the mismatching brackets in comments. Considering this, it's not worth it fixing the bracket matching for now as that would mean creating a more sophisticated textmate grammar.

PR #403, opened 2025-11-18
std: implement accurate F32 parsing

nilscrm:
Implements slow but accurate F32 parsing with `Nat`. Even if we implement a faster algorithm in the future (e.g. Eisel-Lemire) a fully precise fall-back is always needed.

### DIFF
--- a/vine/std/numeric/F32/parse.vi
+++ b/vine/std/numeric/F32/parse.vi
@@ -142,7 +142,7 @@ mod Rational {
 
       // Subnormal float
       if exp == -150 {
-        // In some edge cases, exp can be -150 but the value still rounds up to 1e-149.
+        // In some edge cases, exp can be -150 but the value still rounds up to 2^-149.
         den <<= 1;
       } else {
         num <<= (exp + +149) as N32;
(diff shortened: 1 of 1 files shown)


tests/snaps/vine/repl/parse_f32.repl.vi
```diff
@@ -135,27 +147,60 @@ let io: IO = <IO>;
...
+> F32::parse("1.401298464324817e-45").unwrap().to_bits().to_hex()  // smallest subnormal
+"0x1"
+
+let io: IO = <IO>;
+> F32::parse("1.1754942e-38").unwrap().to_bits().to_hex()  // largest subnormal
+"0x7fffff"
+
+let io: IO = <IO>;
+> F32::parse("1.0000000596046448").unwrap().to_bits().to_hex()  // halfway between 1 and next float
+"0x3f800001"
```
tjjfvi: Shouldn't this be 1, from round-to-even?

vine/std/numeric/F32/parse.vi
```diff
@@ -0,0 +1,178 @@
...
+      if mantissa == 0 {
+        return 0.0;
+      }
+
+      // rounding can push a subnormal to the smallest normal
+      if mantissa >= (1 << 23) {
+        F32::from_parts(false, 1, 0)
+      } else {
+        F32::from_parts(false, 0, mantissa)
+      }
```
tjjfvi: this check is actually unnecessary because an overflow in the mantissa will set the lowest bit of the exponent

tjjfvi (approved):

PR #406, opened 2025-11-19
std: add Heap

nilscrm:


vine/std/data/Heap.vi
```diff
@@ -0,0 +1,118 @@
...
+  pub impl fork[T+]: Fork[Node[T]] {
+    fn fork(&Node[T](value, left, right)) -> Node[T] {
+      Node(value, left.fork(), right.fork())
+    }
+  }
+
+  #[become(unsafe::erase)]
+  pub impl drop[T?]: Drop[Node[T]] {
+    fn drop(Node[T](value, left, right)) {}
+  }
```
tjjfvi: These are equivalent to the basic impls so `struct*` can be used instead

vine/std/data/Heap.vi
```diff
@@ -0,0 +1,148 @@
...
+      swap(&self, &other);
+    }
+    let Heap(_, Node(value, left, right)) = self;
+    let merged = right.merge(other);
+    let (new_left, new_right) = if left.rank() >= merged.rank() {
+      (left, merged)
+    } else {
+      (merged, left)
+    };
+    Heap(new_right.rank() + 1, Node(value, new_left, new_right))
```
tjjfvi:
```suggestion
    let &Heap(rank, Node(value, left, right)) = &self;
    right = right.merge(other);
    if right.rank() >= left.rank() {
      swap(&left, &right);
    }
    rank = right.rank() + 1;
    self
```

tjjfvi: Maybe it would also be nicer for `merge` to be `fn(&Heap, Heap)`?

PR #408, opened 2025-11-20
std: use mod generics for Iterator

nilscrm:


tjjfvi (approved):
not sure why I didn't do this in the first place...

PR #412, opened 2025-11-24
vine: require content type in enum variants

tjjfvi:
Requires content in enum variants, so instead of `Some(123)` and `None`, it is now `Some(123)` and `None()`. This:
- removes a meaningless distinction
- addresses an ambiguity in the grammar, where an identifier in a pattern could either be a variable or refer to a content-less enum variant.

Additionally, this PR allows deduplicating parens when declaring a struct/enum variant with tuple content, so `struct N64((N32, N32))` can now just be `struct N64(N32, N32)`, mirroring the pattern and value syntax sugar.

### DIFF
--- a/tests/snaps/vine/repl/misc.repl.vi
+++ b/tests/snaps/vine/repl/misc.repl.vi
@@ -460,6 +460,6 @@ hi
 
 let io: IO = <IO>;
 > Some(123) is None
-error input:1:14 - expected data subpattern
+error input:1:14 - expected content subpattern
 
 let io: IO = <IO>;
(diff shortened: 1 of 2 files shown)


tjjfvi:
> I see the grammar ambiguity but I think it's extremely uncommon as it is discouraged by naming convention.

The biggest problem with the grammar ambiguity is that if you write something like
```rs
match foo {
  Nome { "None" }
  Some(x) { x }
}
```
it will always output `"None"`, because `None` was typoed as `Nome` and rather than getting an error, it silently became a variable pattern.

nilscrm:
> The biggest problem with the grammar ambiguity is that if you write something like
> 
> ```rust
> match foo {
>   Nome { "None" }
>   Some(x) { x }
> }
> ```
> 
> it will always output `"None"`, because `None` was typoed as `Nome` and rather than getting an error, it silently became a variable pattern.

Oh yeah this is a pretty bad one. But I would say this gets even worse with the new design as now the same error occurs even for:
```rs
match foo {
  None { "None" }
  Some(x) { x }
};
```

PR #415, opened 2025-11-28
vine: error detection

tjjfvi:
The runtime will now error if:
- `freed < allocated` (indicating the program created a vicious circle)
- the IO handle is not returned in the reference from `main` (indicating that something erased it, or some other similar issue)

This means that we can now detect if a program encounters an error, even outside of debug mode.

### DIFF
--- a/cli/src/common.rs
+++ b/cli/src/common.rs
@@ -35,7 +35,7 @@ pub struct RunArgs {
 }
 
 impl RunArgs {
-  pub fn run(self, nets: Nets) {
+  pub fn run(self, nets: Nets, debug_hint: bool) {
     let mut host = &mut Host::default();
     let heap = Heap::new();
     let mut extrinsics = Extrinsics::default();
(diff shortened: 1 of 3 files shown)


vine/src/components/emitter.rs
```diff
@@ -294,3 +294,24 @@ impl<'a> Emitter<'a> {
     (Tree::Var(label.clone()), Tree::Var(label))
   }
 }
+
+pub(crate) fn main_net(debug: bool, main: Tree) -> Net {
+  if debug {
+    main_net_debug(main)
+  } else {
+    Net {
+      root: Tree::n_ary("x", [Tree::Var("io0".into()), Tree::Var("io1".into())]),
```
nilscrm: Instead of destructing the `x` node and then combining the wires again into a `ref` node couldn't the ivm just link the main global directly to an io `ref` node?

nilscrm (approved):

vine/src/components/emitter.rs
nilscrm: For me both works, thought it might make sense as other references are also passed as `ref` nodes.

PR #416, opened 2025-11-28
vine: use keywords for all control flow

tjjfvi:
Ensures all control flow is denoted with a keyword. In particular:
- `a && b` is now `a and b`
- `a || b` is now `a or b`
- `a => b` (the implies operator) is now `a impl b`
- `result?` is now `result.try`

(Note that `!a` is still `!a`, as opposed to `not a`, as it is not control flow, and the precedence of `!a and b` is much clearer than that of `not a and b`.)

### DIFF
--- a/vine/src/components/lexer.rs
+++ b/vine/src/components/lexer.rs
@@ -76,9 +76,9 @@ pub enum Token {
   #[token(">=")]
   Ge,
   #[token("->")]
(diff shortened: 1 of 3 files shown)


vine/src/components/lexer.rs
nilscrm: The `=>` token is not needed anymore.

PR #431, opened 2025-12-03
vine: warnings & pretty diagnostics

tjjfvi:
Adds the ability to issue warnings, and two basic warnings: unused variable and unused item.

Since I was touching diagnostic printing anyway, I made it a bit prettier:
```rs
pub fn main(&io: &IO) {
  iyo.println("Hello, world!");
}
```
image

### DIFF
--- a/tests/snaps/vine/fail/atypical.txt
+++ b/tests/snaps/vine/fail/atypical.txt
@@ -1,73 +1,290 @@
-error tests/programs/fail/atypical.vi:3:19 - no type associated with `#atypical::foo`
-error tests/programs/fail/atypical.vi:3:24 - types in item signatures cannot be elided
-error tests/programs/fail/atypical.vi:5:12 - types in item signatures cannot be elided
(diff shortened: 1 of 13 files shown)


lsp/src/lib.rs
```diff
@@ -41,36 +42,31 @@ impl Backend {
...
-      Err(diags) => diags,
-    };
+    _ = lsp.compiler.compile(());
     eprintln!("compiled in {:?}", start.elapsed());
 
-    diags.sort_by_key(|d| Some(d.span()?.file));
-    let mut diags = diags.into_iter().peekable();
-    while diags.peek().is_some_and(|x| x.span().is_none()) {
-      diags.next();
+    let mut diags_by_file = HashMap::<FileId, (Vec<&Diag>, Vec<&Diag>)>::new();
```
enricozb: should we maybe name `(Vec<&Diag>, Vec<&Diag>)` so we know which are errors and which are warnings?

Issue #118, opened 2024-12-19
vine: basic test system


tjjfvi:
```rs
#[test]
fn foo(&io: &IO) {
  assert 1 < 2 else {
    debug::error("math is broken")
  }
}
```

- functions can be marked as `#[test]`s
- tests run in debug mode
- `debug::error` causes the test to fail
- `vine test` runs all tests defined in the main module, reporting successes & failures
  - support filtering by path

PR #435, opened 2025-12-04
vine: `#[test]` and `#[cfg(test)]` attributes

enricozb:
initial work for: https://github.com/VineLang/vine/issues/118

- [x] `#[test]` attribute can be attached to `fn(&io: &IO) -> ()` items
- [x] `vine test <main.vi>` will print out the paths of all `#[test]` fns in that module
- [x] `#[cfg(test)]` conditionally compiles code when run under `vine test`
- [x] `vine test <main.vi> <path>` runs the `<path>` in `<main.vi>` as if it was `main`

what remains:
- [ ] `debug::error` causes the test to fail

### DIFF
--- a/vine/src/components/resolver.rs
+++ b/vine/src/components/resolver.rs
@@ -607,14 +607,8 @@ impl<'a> Resolver<'a> {
   }
 
   pub(crate) fn resolve_test_fn(&mut self, concrete_fn_id: ConcreteFnId) {
-    match self.expect_entrypoint_sig(concrete_fn_id) {
-      Err(Diag::GenericEntrypoint { span }) => {
-        self.diags.error(Diag::GenericEntrypoint { span });
-      }
-      Err(diag) => {
-        self.diags.error(diag);
-      }
-      _ => {}
+    if let Err(diag) = self.expect_entrypoint_sig(concrete_fn_id) {
+      self.diags.error(diag);
     }
   }
 }


vine/src/structures/diag.rs
```diff
@@ -293,6 +295,8 @@ diags! {
     ["unused variable"]
   UnusedItem
     ["unused item"]
+  BadTestAttr
+    ["the `#[test]` attribute can only be applied to an fn"]
```
tjjfvi: Can you change this and `BadFramelessAttr` to say `a function`? I don't know why I wrote it as `an fn`...

vine/src/features/cfg.rs
```diff
@@ -14,11 +14,33 @@ use crate::{
...
+    }
+  }
+
+  pub fn test(&self) -> bool {
+    match self.values.get(&Ident("test".into())) {
+      Some(ConfigValue::Bool(test)) => *test,
+      _ => false,
+    }
+  }
+}
```
tjjfvi: I think for the `_ =>` cases an `unreachable!()` might be nicer

tests/tests.rs
```diff
@@ -207,6 +211,15 @@ fn test_vi_repl(t: &mut DynTester, path: &'static str) {
   });
 }
 
+fn test_vi_test(t: &mut DynTester, path: &'static str, test: &'static str) {
+  let name = Path::file_stem(path.as_ref()).unwrap().to_str().unwrap();
+  t.test("test", move || {
+    let (stdout, stderr) = exec(VINE, &["test", "--no-stats", path, test], &[], true);
+    assert_empty_stderr(&stderr);
+    test_snapshot(&["vine", "test", &format!("{name}.{test}.txt")], &stdout);
```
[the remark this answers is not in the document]
enricozb: Yeah, I wasn't sure what the right thing to do here was. I didn't _really_ want to introduce a new `test_vi_*` function, but i didn't see another way to add tests for `#[test]`.

Once this attribute is merged and usable, we should probably have CI tests for root using `#[test]`, which would implicitly test this.

But having an explicit test also makes sense. Maybe this subdir just won't have much added to it.

tjjfvi (approved):

PR #438, opened 2025-12-05
std: rename range.vi -> Range.vi

nilscrm:
This allows path collapsing when importing `Range` and is more aligned with the current naming convention.


enricozb:
what kind of path collapsing are you referring to? the modified import in `List.vi` looks the same as before (other than the casing).

nilscrm:
> what kind of path collapsing are you referring to? the modified import in `List.vi` looks the same as before (other than the casing).

The import didn't change since we're also importing `Bound` but if that's not the case one can now use `#root::ops::Range` instead of `#root::ops::range::Range`

enricozb:
I think part of the confusion is that `range.vi` is the only(?) file exporting more than one struct/trait (`Bound` and `Range`), but also shares a name with one of those structs/traits. So perhaps we should consider moving `Bound` to its own file?

enricozb (approved):

tjjfvi:
Thanks, I've been meaning to do this.

nilscrm:
> I think part of the confusion is that `range.vi` is the only(?) file exporting more than one struct/trait (`Bound` and `Range`), but also shares a name with one of those structs/traits. So perhaps we should consider moving `Bound` to its own file?

It's equivalent to having `Bound` inside of `Range`. E.g. importing `List` and its iterator also looks like `use #root::data::List::{List, Iter}`

Issue #219, opened 2025-04-27
vine: ensure everything is nameable


tjjfvi:
#212 is introducing some unnameable ("Voldemort") types & implementations:
- the type of a `fn` item / `let fn` / `fn` expression
- the automatic implementations of the `fn`, `Fork`, and `Drop` traits for the above

All types should be nameable, so they can e.g. be returned in functions.

All implementations should be nameable so they can be explicitly specified when ambiguity arises.

PR #447, opened 2025-12-09
root: add Map::keys and Map::values

nilscrm:


root/data/Map.vi
```diff
@@ -127,6 +127,34 @@ pub mod Map[K, V; Ord[K]] {
     }
   }
 
+  pub fn .keys[...](&self: &Map[K, V]) -> Iterator::Map[IterRef[K, V], Key] {
+    self.iter_ref().map(Key())
```
enricozb: Is it not possible to pass in an fn here instead of a Key struct?

same question for Value.

nilscrm: Unfortunately not, since then you can't refer to the function type of that lambda in the signature. The same thing happens in `Iterator::inspect`.

enricozb: Ah I guess you'd have to name the type of the closure

tjjfvi (approved):

root/data/Map.vi
tjjfvi: #219

PR #448, opened 2025-12-09
root: implement Collect for Set

nilscrm:


tjjfvi:
could be nice to add this for `Map` as well

nilscrm:
> the `Drop[T]` parameter is before the `Iterator[I, T]` parameter

Makes sense and I guess with the way the `Collect` trait is defined it cannot be rearranged.

root/data/Set.vi
```diff
@@ -276,6 +276,16 @@ pub mod Set[T; Ord[T]] {
     }
   }
 
+  pub impl collect[T?; Ord[T]]: Collect[Set[T], T] {
+    fn collect[I; Iterator[I, T]](iter_ref: I) -> Set[T] {
+      let set = Set::empty;
+      for value in iter_ref {
```
tjjfvi:
```suggestion
    fn collect[I; Iterator[I, T]](iter: I) -> Set[T] {
      let set = Set::empty;
      for value in iter {
```

PR #453, opened 2025-12-11
root: add Iterator::chunks

enricozb:


enricozb:
> I believe currently `[1, 2, 3, 4].chunks(2)` will return `[1, 2], [3, 4], []`

nice catch, pushed a fix.

tjjfvi (approved):

PR #465, opened 2025-12-18
root: implement Map::map_values

nilscrm:


root/data/Map.vi
```diff
@@ -98,6 +98,10 @@ pub mod Map[K, V; Ord[K]] {
     self!.remove_max()
   }
 
+  pub fn .map_values[... U, F*; fn F(V) -> U](self: Map[K, V], f: F) -> Map[K, U] {
```
tjjfvi: I think it would be useful for this to be `fn F(&K, V) -> U`

PR #472, opened 2025-12-27
vine: remove unused keywords

tjjfvi:


vine/src/components/lexer.rs
```diff
@@ -70,13 +69,10 @@ pub enum Token {
   Impl,
   Match,
   Let,
-  Dyn,
```
nilscrm: Do we also want to remove `Dyn` if we know we want to use it very soon?

tjjfvi: I think it's nicer to just remove it, even if we add it back soon. We might come up with a different keyword, and I think in general it's nice not to have unused keywords.

nilscrm: Fair I guess we don't care about breaking changes at all for now

nilscrm (approved):

PR #473, opened 2025-12-29
ivy: write erasers on invalid extrinsic operations

enricozb:
Write erasers to outgoing wires, and set an IVM flag, when:
- copying a non-copyable extrinsic type
- providing an extrinsic value of an unexpected type to an extrinsic function

### DIFF
--- a/root/numeric/N64.vi
+++ b/root/numeric/N64.vi
@@ -125,20 +125,12 @@ pub mod N64 {
 
   pub impl div: Div[N64, N64, N64] {
     fn div(a: N64, d: N64) -> N64 {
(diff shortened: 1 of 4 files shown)


tests/programs/invalid_extrinsics.vi
```diff
@@ -0,0 +1,9 @@
+
+pub fn main(&_io: &IO) {
+  let f = 1.0;
+  let n = 1;
+  inline_ivy! (n <- n, f <- f) -> () {
+    out
+    f = @n32_add(n out)
+  };
+}
```
enricozb: Not sure if we want to put this inside of an ivy test folder instead.

PR #476, opened 2025-12-30
vine: anonymous impls

### LINKED ISSUE #210: vine: optional impl names
Associate unnamed implementations with the parent module. Allow one name to refer to multiple implementations as long as they are of distinct traits.

Impls will still always be able to be explicitly specified, but most impls will now no longer need to be named.

tjjfvi:
Resolves #210

### DIFF
--- a/vine/src/components/finder/candidates.rs
+++ b/vine/src/components/finder/candidates.rs
@@ -292,15 +292,9 @@ impl<'a> CandidateSetBuilder<'a> {
   }
 
   fn add_direct_candidate(&self, set: &mut CandidateSetHandle, (def_id, level): Node) {
-    if level == Level::WithinAncestors {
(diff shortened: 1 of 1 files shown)


vine/src/components/finder/candidates.rs
```diff
@@ -283,13 +292,16 @@ impl<'a> CandidateSetBuilder<'a> {
   }
 
   fn add_direct_candidate(&self, set: &mut CandidateSetHandle, (def_id, level): Node) {
-    if level != Level::Direct {
+    if level == Level::WithinAncestors {
       return;
     }
```
enricozb: Do we want to instead, match on exactly which levels we care about?
```
if matches!(level, Level::Direct | Level::Within) { .. }
```

Issue #475, opened 2025-12-29
vine: dynamic types


tjjfvi:
Add dynamic types (also known in other contexts as existential types).

Dynamic types are written `dyn[<generics>] <type>`, e.g. `dyn[T; Show[T]] &T`. This type, for example, is similar to Rust's `&dyn Show`.

Similarly, they can be used to implement function pointers, e.g. `dyn[F; fn F(A, B) -> C] F` would be similar to Rust's `Box<dyn FnOnce(A, B) -> C>`.

The dynamic expression, written `dyn[<generics>] <expr>` constructs a dynamic type. For example, `dyn[N32; N32::show] &123` would be an expression of type `dyn[T; Show[T]] &T`. If the contextual type is known, the generics could be inferred as usual, so this could also be written `dyn &123`.

The dynamic pattern, written `dyn[<generics>] <pattern>` destructs a dynamic type. For example:
```rs
impl show_dyn: Show[dyn[T; Show[T]] &T] {
  fn show(&dyn &value: &dyn[T; Show[T]] T) -> Show {
    value.show()
  }
}
```

These types will likely be verbose to work with at first, but they are a neccesary primitive, and support can be improved over time.

Potential sugar includes:
- `dyn TraitName` for `dyn[T; TraitName[T]] T` for traits with a single generic parameter
- `dyn fn* (A, B) -> C` for `dyn[F*; fn F(A, B) -> C] F` (and similarly with other flex annotations)
- `dyn &TraitName` for `dyn[T; TraitName[T]] &T`
- `dyn &fn* (A, B) -> C` for `dyn[F*; fn F(A, B) -> C] &F`
- using the unwrap operator (`!`) to access the content of a dynamic type

I'm not sure what the best mechanism for automatically implementing traits for dynamic types would be. Initially, `Fork` and `Drop` could be implemented automatically by the compiler when the content is always forkable/dropable.

tjjfvi:
I've figured out a very general and powerful system for implementing traits for dynamic types:

```rs
trait DynCast[A, B] {
  fn cast(a: A) -> B;
}
```

The compiler would automatically implement the `DynCast` trait for every pair of dynamic types `A` and `B` such that `A` is essentially a subtype of `B`.

This seemingly very simple trait allows for very powerful reflection on dynamic types. Perhaps the simplest example is `Drop`:
```rs
impl drop_dyn[D; DynCast[D, dyn[T?] T]]: Drop[D] { ... }
```
[...]

PR #481, opened 2025-12-31
vine: conditional types

### LINKED ISSUE #480: vine: conditional types
Add conditional types, written
```rs
type Foo = if const path::to::some::constant { Bar } else { Baz };
```
These have the semantics that, if the constant is `true`, the type is equivalent to `Bar`, and otherwise it is equivalent to `Baz`.

However, at type-checking time, the value of the constant is not known, and so both branches need to be handled.

This is done with the `if const` expression, written
```rs
let foo = if const path::to::some::constant { bar } else { baz };
foo // if const path::to::some::constant { Bar } else { Baz }
```

This, combined with #479, covers the majority of uses of conditional compilation, whilst being a lot simpler/saner (e.g. type checking happens once, independently of the config; vs with conditional compilation, type checking must occur with every combination of the compilation options).

tjjfvi:
Resolves #480 

Conditional types are currently somewhat annoying to work with -- using them safely requires passing them through the output of an `if const`, which is awkward for consumption -- but this can be improved in the future. Particularly, the biggest missing piece is being able to cast `if const foo { A } else { B }` to & from `A`/`B` in the true/false arms. Supporting this requires a mechanism also required by #475.

`debug::WithState` has been updated to use a conditional types. This removes the last use of conditional compilation (`#[cfg(...)]`) from the repo, and so this PR also removes support for conditional compulation.

### DIFF
--- a/vine/src/components/finder.rs
+++ b/vine/src/components/finder.rs
@@ -349,26 +349,19 @@ impl<'a> Finder<'a> {
       ImplType::Trait(trait_id, type_params) => {
         if Some(*trait_id) == self.chart.builtins.fn_ {
           self.find_auto_impls_fn(type_params, types, found)?;
-        }
-        if Some(*trait_id) == self.chart.builtins.fork {
+        } else if Some(*trait_id) == self.chart.builtins.fork {
           self.find_auto_impls_fork(type_params, types, found);
-        }
-        if Some(*trait_id) == self.chart.builtins.drop {
+        } else if Some(*trait_id) == self.chart.builtins.drop {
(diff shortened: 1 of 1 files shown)


vine/src/components/finder.rs
```diff
@@ -368,6 +368,9 @@ impl<'a> Finder<'a> {
         if Some(*trait_id) == self.chart.builtins.enum_ {
           self.find_auto_impls_enum(type_params, types, found);
         }
+        if Some(*trait_id) == self.chart.builtins.if_const {
```
enricozb: should this be a chain of else-if's?

PR #485, opened 2026-01-02
ci: integration testing

tjjfvi:
Adds integration testing, so that we can ensure other repos using Vine are kept up-to-date with Vine changes.

Going forward:
- PRs will be made against the `dev` branch, instead of `main`.
  - If a PR is opened against `main`, it will automatically be changed to `dev`.
- Once a commit is made to `dev`, the integration tests will be run.
- Whenever the integration tests pass, `main` will be updated to that commit.
- If integration tests fail, #484 will be (re)opened and assigned to the person who pushed the failing commit.
- Once the other repos are fixed, the integration tests can be rerun, and if they succeed, `main` will be updated.

Example integration tests run: https://github.com/VineLang/vine/actions/runs/20666244162/job/59345097604

Currently, only [Liana](https://github.com/nilscrm/liana) is included in the integration tests. More repos can be added in the future. For a repo to be added, I would like it to have a Nix flake such that `nix flake check` suffices to run the relevant tests. cc @nilscrm

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -36,7 +36,7 @@ jobs:
 
   notify:
     needs: [mainline]
-    if: ${{ github.ref_name == 'dev' }}
+    if: ${{ always() && github.ref_name == 'dev' }}
     runs-on: ubuntu-latest
     permissions:
       contents: read


.github/workflows/integration.yml
```diff
@@ -0,0 +1,62 @@
...
+    timeout-minutes: 10
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - run: |
+          git fetch origin main
+          if [[ -z `git branch origin/main --contains @` ]]; then
```
nilscrm: Should we use `git merge-base --is-ancestor HEAD origin/main;` instead for robustness and to make sure this is a fast-farward.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,62 @@
...
+        run: |
+          gh issue reopen 484
+          gh issue edit 484 --add-assignee ${{ github.event.sender.login }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+      - name: close issue
+        if: ${{ needs.mainline.result == 'success' }}
+        run: |
+          git fetch origin main
+          if [[ `git rev-parse dev` == `git rev-parse origin/main` ]]; then
```
nilscrm: I think CI is usually is detached head and I'm not sure it has a local `dev` branch
```suggestion
          if [[ `git rev-parse origin/dev` == `git rev-parse origin/main` ]]; then
```

.github/workflows/integration.yml
```diff
@@ -0,0 +1,62 @@
...
+          fetch-depth: 0
+      - run: |
+          git fetch origin main
+          if [[ -z `git branch origin/main --contains @` ]]; then
+            git push origin `git rev-parse @`:main
+          fi
+
+  notify:
+    needs: [mainline]
+    if: ${{ github.ref_name == 'dev' }}
```
nilscrm: I think this means that `notify` is skipped if `mainline` is skipped or failed, so it won't run exactly when we want the notification to happen, see [here](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-not-requiring-successful-dependent-jobs) for the fix.
```suggestion
    if: ${{ always() && github.ref_name == 'dev' }}
```

.github/workflows/integration.yml
```diff
@@ -0,0 +1,62 @@
...
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - run: |
+          git fetch origin main
+          if [[ -z `git branch origin/main --contains @` ]]; then
+            git push origin `git rev-parse @`:main
```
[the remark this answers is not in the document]
tjjfvi: I can set up the branch protection so that it does not

.github/workflows/integration.yml
```diff
@@ -0,0 +1,62 @@
...
+          fi
```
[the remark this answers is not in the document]
tjjfvi: The else branch of this condition only runs when `@` is already a part of `main`, so there is no need to push or fail. This could happen if, for example, main was updated to a later commit before the integration tests finished for an earlier commit.

PR #486, opened 2026-01-03
ci: add sat-vi and page-rank to integration tests

nilscrm:
Example run [here](https://github.com/VineLang/vine/actions/runs/20748336012)

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -14,7 +14,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


tjjfvi:
These seemingly don't appear as checks, meaning they can't be required to push to `main`. Perhaps you can add a no-op check that depends on the whole matrix that can be used? (That could also be used as the dependent of `notify` instead of `mainline`)

tjjfvi:
E.g. have `integration-test` with the matrix, `integration-tests` depend on the whole matrix, and `mainline` and `notify` depending on `integration-tests`.

tjjfvi:
The `repo` field in the matrix also seems unnecessary.

PR #492, opened 2026-01-06
vine: unsafe enums

### LINKED ISSUE #488: vine: unions
Add unions, which are like enums, except that they do not store their variant; their variant must be tracked out-of-band. These are similar to Rust unions.

Currently, the standard library has types such as `Array::Node` *should* be unions, but are currently represented as a struct for one of the variants, and are transmuted to other variants when appropriate. This has poor ergonomics, and will ultimately be UB, so these types will be transitioned to be `union`s instead.

For example, a tree type whose shape is stored out-of-band could be represented as follows:
```rs
union Node[T] {
  leaf(T),
  branch(Node[T], Node[T]),
}
```
Union variant names are conventionally written in `snake_case`, distinguishing them visually from enum variants.

Because the variant of a union is not stored inline, all operations using them are [unsafe](https://github.com/vineLang/vine/issues/487).

Unions support constructor syntax (`Node::leaf(123)`). This works for values/places/spaces, like struct constructors. Union constructor patterns are irrefutable.

The content of a union can also be accessed with field syntax, specifying which variant the union is inhabited (e.g. `node.leaf` or `node.branch`). (This is the other reason union variants are written in `snake_case`.)

tjjfvi:
Resolves #488 

Inside the compiler, unsafe enums are referred to as unions, to more succinctly distinguish them from safe enums.

Depends on #491

### DIFF
--- a/tests/programs/fail/unsafe_.vi
+++ b/tests/programs/fail/unsafe_.vi
@@ -1,4 +1,6 @@
 
+use #root::logical::Nullable;
+
 pub unsafe fn main(_: &IO) {} // error
 
 pub unsafe const foo: () = ();
@@ -7,6 +9,10 @@ pub unsafe fn bar() {
   foo; // ok
   bar(); // ok
   inline_ivy! () -> () { _ } // ok
+  let nullable = Nullable::null(); // ok
+  &nullable.null; // ok
+  let Nullable::null() = nullable; // ok
+
   safe foo; // warning
 }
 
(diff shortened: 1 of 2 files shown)


vine/src/features/composite.rs
```diff
@@ -318,62 +318,82 @@ impl Resolver<'_> {
...
     index: usize,
   ) -> Result<TirExpr, Diag> {
-    let tuple = self.resolve_expr(tuple);
-    let (ty, fields) = self.tuple_field(span, tuple.ty, index).ok_or_else(|| {
-      Diag::MissingTupleField { span, ty: self.types.show(self.chart, tuple.ty), i: index }
-    })?;
-    Ok(TirExpr::new(span, ty, TirExprKind::Field(tuple, index, fields)))
+    let expr = self.resolve_expr(expr);
+    let ty = expr.ty;
+    self._resolve_expr_tuple_field(span, expr, index).ok_or_else(|| Diag::MissingTupleField {
```
nilscrm: Instead of having the `ok_or_else` I think it is cleaner to just have `_resolve_expr_tuple_field` return a `Result<(Type, Vec<Type>), Diag>`

nilscrm (approved):

Issue #479, opened 2025-12-31
vine: configurable constants


tjjfvi:
Allow certain constants to be marked as `#[configurable]`.
```rs
#[configurable]
const max_length: N32 = 100;
```
Configurable constants can have their value overridden when compiling:
```
vine run foo.vi --config max_length 1000
```
For now, configurable constants would be restricted to `Bool`, `N32`, and `String`, but this could be generalized in the future.

PR #494, opened 2026-01-19
tests: rewrite in nix

### LINKED ISSUE #49: snapshots should detect when there snapshots with no corresponding test

tjjfvi:
Rewrites the tests & ci in nix. Resolves #49.

There's a lot of churn here, so I highly recommend reviewing by-commit.

The CI now simply runs `nix flake check` on a self-hosted runner. Because of caching, this almost never has to run the full test suite, and is thus much faster than the previous CI.

Additionally, one can run all checks locally with `nix flake check`.

It will now be much easier to add various ad-hoc tests (as needed, for example, to test #479).

Two changes to the code were made as part of this:
- A `--no-perf` flag was added to the ivy cli; this disables the printing of non-deterministic statistics, allowing the statistics to be snapshotted.
- Instead of printing file paths relative to the current directory, the compiler now prints a 'virtual' file path, starting from the top-level source. So for example, `vine run vine/examples/hello_world.vi` will now report errors as being from `hello_world.vi` instead of the full `vine/examples/hello_world.vi` path, making it independent of where the source is located or where the cli is invoked.

I've also changed the integration tests to only run Liana's example check and not the full test suite because the test suite is extremely slow.

### DIFF
--- a/tests/tests.nix
+++ b/tests/tests.nix
@@ -7,11 +7,8 @@
 }:
 let
   inherit (builtins)
-    map
     filter
     attrNames
-    dirOf
-    baseNameOf
     concatLists
     attrValues
     concatStringsSep
@@ -340,4 +337,4 @@ let
 
   all = merge ([ self ] ++ concatLists (attrValues tests));
 in
-builtins.removeAttrs all [ "snaps" ]
+removeAttrs all [ "snaps" ]


flake.nix
```diff
@@ -40,93 +40,147 @@
...
 
-          packages = {
-            default = vine;
-            inherit vine;
-          };
+        packages.default = self.cli.packages.vine;
 
-          devShells.default = craneLib.devShell {
-            name = "vine-dev";
+        checks.fmt =
```
tjjfvi: this thing is huge and probably shouldn't be in the flake directly but I'm not sure where else to put it

vine/src/components/loader.rs
```diff
@@ -148,11 +155,62 @@ impl<F: FS> VisitMut<'_> for LoadChildren<'_, '_, F> {
...
+  }
+
+  fn child_file(&mut self, path: &Self::Path, name: &Ident) -> Self::Path {
+    (*self).child_file(path, name)
+  }
+
+  fn read_file(&mut self, path: &Self::Path) -> Option<String> {
+    (*self).read_file(path)
+  }
+}
```
nilscrm: I don't think this implementation is used.

tests/tests.nix
```diff
@@ -0,0 +1,294 @@
...
+
+  just =
+    src:
+    let
+      dir = pkgs.lib.fileset.toSource {
+        root = dirOf src;
+        fileset = src;
+      };
+    in
+    "${dir}/${baseNameOf src}";
```
nilscrm: We can get around copying each individual file but using link farms.
```suggestion
  directive =
    keyword: default: src:
    let
      asDirFile = "${src}/${baseNameOf src}.vi";
      content =
        if builtins.pathExists asDirFile then builtins.readFile asDirFile else builtins.readFile src;
      match = builtins.match "(.*\n)?// @${keyword} ?([^\n]*)\n.*" content;
    in
    if match == null then default else builtins.elemAt match 1;

  libs =
    path:
    let
      libs = directive "libs" "" path;
    in
    if libs == "" then
      ""
    else
      concatStringsSep " " (map (x: "--lib ${just ../${x}}") (splitString " " libs));

  just =
    src:
    let
      dir = pkgs.linkFarm "vine-src" [
        {
          name = baseNameOf src;
          path = src;
        }
      ];
    in
    dir + "/${baseNameOf src}";
```

tests/tests.nix
```diff
@@ -0,0 +1,335 @@
...
+                pkgs.runCommand "sysroot-cargoLock"
+                  {
+                    nativeBuildInputs = [ rustToolchain ];
+                  }
+                  ''
+                    cp `rustc --print sysroot`/lib/rustlib/src/rust/library/Cargo.lock $out
+                  '';
+              dontFixup = true;
+              cargoArtifacts = null;
+            });
```
nilscrm:
```suggestion
            MIRI_SYSROOT = craneLib.mkCargoDerivation {
              pname = "miri-sysroot";
              version = "0.0.0";
              dontUnpack = true;
              buildPhaseCargoCommand = ''
                MIRI_SYSROOT=$out cargo miri setup
              '';
              cargoLock =
                pkgs.runCommand "sysroot-cargoLock"
                  {
                    nativeBuildInputs = [ rustToolchain ];
                  }
                  ''
                    cp `rustc --print sysroot`/lib/rustlib/src/rust/library/Cargo.lock $out
                  '';
              dontFixup = true;
              cargoArtifacts = null;
            };
```

tests/tests.nix
```diff
@@ -0,0 +1,335 @@
...
+  inherit (builtins)
+    map
+    filter
+    attrNames
+    dirOf
+    baseNameOf
+    concatLists
+    attrValues
+    concatStringsSep
+    ;
```
nilscrm:
```suggestion
  inherit (builtins)
    filter
    attrNames
    concatStringsSep
    attrValues
    concatLists
    ;
```

tests/tests.nix
```diff
@@ -0,0 +1,335 @@
...
+
+  self.apps.snaps = flake-utils.lib.mkApp {
+    drv = pkgs.writeShellScriptBin "snaps" ''
+      ${pkgs.nushell}/bin/nu ${./snaps.nu} ${snaps} ./tests/snaps --write
+    '';
+  };
+
+  all = merge ([ self ] ++ concatLists (attrValues tests));
+in
+builtins.removeAttrs all [ "snaps" ]
```
nilscrm:
```suggestion
removeAttrs all [ "snaps" ]
```

tests/tests.nix
```diff
@@ -0,0 +1,294 @@
...
+      key ? "programs",
+    }:
+    let
+      name = sanitize path;
+      vi = just path;
+      buildArgs = directive "build" "" config;
+      runArgs = directive "run" "" config;
+      input = ./input/${directive "input" "empty" config};
+      ext = directive "output" "txt" config;
+      fail = directive "fail" null config != null;
```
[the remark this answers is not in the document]
tjjfvi: does nix not memoize functions?

tests/tests.nix
nilscrm: Actually not sure about the documentation but the [`nixd`](https://github.com/nix-community/nixd) LSP picks up on these.

tests/tests.nix
nilscrm: Agree that githubs diff looks ugly. The reason I still put them together is that if you change `just` to creating symlinks, `directive` can't handle symlinks to directories as it makes an `file_type == "directory"` check which fails if it is a symlink to a directory.

That being said I originally thought the link farm would be more efficient as you otherwise end up with duplicate input (one store path with all test files and then one for each individual test file). But looking into it I think I was wrong on this and nix does not create the directory with all test programs.

vine/src/components/loader.rs
```diff
@@ -148,11 +155,62 @@ impl<F: FS> VisitMut<'_> for LoadChildren<'_, '_, F> {
...
+
 pub trait FS {
```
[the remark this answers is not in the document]
nilscrm: That's fair, I'm okay with keeping it for now rather than reimplementing it again in the future.

vine/src/components/loader.rs
```diff
@@ -148,11 +155,62 @@ impl<F: FS> VisitMut<'_> for LoadChildren<'_, '_, F> {
   }
 }
 
+struct DisplayFS<F>(F);
```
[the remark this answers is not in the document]
nilscrm: see [this](https://github.com/VineLang/vine/pull/494#discussion_r2779789572)

cli/src/vine_cli.rs
```diff
@@ -380,15 +380,46 @@ impl VineReplCommand {
 }
 
 #[derive(Debug, Args)]
-pub struct VineFmtCommand {}
+pub struct VineFmtCommand {
+  files: Option<Vec<PathBuf>>,
```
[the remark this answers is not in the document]
tjjfvi: yeah, I just thought it was nicer to check if the files were none than checking if the vec was empty

PR #495, opened 2026-01-20
root: add `File` type

enricozb:
adds the `File` type and basic functions `open` and `.read_bytes`. Example usage looks like:

```rust
use #root::IO::File;

pub fn main(&io: &IO) {
  let file = File::open(&io, "Cargo.toml").assume();
  let bytes = file.read_bytes().assume();
  let content = String(bytes.map(fn*(b: N32) { b as Char }));

  io.println("{content}");

  file.close(&io);
}
```

### DIFF
--- a/ivm/src/ext.rs
+++ b/ivm/src/ext.rs
@@ -9,7 +9,7 @@ use core::{
   ops::{Deref, DerefMut},
 };
 
-use std::fs::File;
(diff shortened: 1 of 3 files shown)


enricozb:
closed in favor of `https://github.com/VineLang/vine/pull/507`

PR #502, opened 2026-01-31
vine: rework loader

### LINKED ISSUE #498: vine: simplify module system with source directories
[In this, I am going to use the term *source* for an input to the vine compiler. Think source file, but not necessarily a file. Sources always define a module, though not all modules correspond to a source. Top-level sources are those that are passed to the vine compiler either as the main source or as library sources, as well as root is `--no-root` is not passed. Non-top-level sources include e.g. `IO.vi`.]

Currently, all Vine sources are files. In files whose path is of the form `.../<name>/<name>.vi`, one can write `mod <child>;` to include either `.../<name>/<child>.vi` or `.../<name>/<child>/<child>.vi` as a source named `<child>`; this item is invalid otherwise.

This somewhat convoluted rule does mostly what we want, but has some issues:
- It is somewhat convoluted, and thus potentially hard to explain.
- It involves inspecting "the path" of a file, which is a very troubling concept. In particular, a file may have multiple different paths. Currently, I believe the Vine compiler uses the canonical, absolute path, with all symlinks followed.
- It is brittle in the face of things like the Nix store, which have the audacity to change components in the path. For example, one cannot include `root` directly in the Nix store, since the name will be changed to `0123456789abcdef-root`, and `0123456789abcdef-root/root.vi` is not of the form `.../root/root.vi`. Instead, one must include the *parent* of `root` into the Nix store, and filter it to just include `root`.

Instead, replace the rules as follows: Vine sources may either be files or directories. Additionally, they have an associated *name*, which must be a valid identifier. The name of a source is equivalent to the name of the module it defines.
- Source directories must have a file at the path `<name>.vi`, which holds the source code that defines the module. In this source code, one may write `mod <child>;`. This looks for either a file named `<child>.vi` or a directory named `<child>` to be the source of the child module.
- Source files are handled as usual. In them, the item `mod <child>;` is invalid.

The shift in thinking here is that, instead of `root/root.vi` being the source, the source is the *directory* `root/`. This greatly simplifies the handling; no more do we ask "what path does this thing have?" but only "what thing is at this path?", which is a much more reasonable question to ask of a file system.

The main user-facing consequence of this change is that instead of running e.g. `vine run foo/foo.vi`, one would run `vine run foo`. I think this is actually a lot cleaner and nicer. (`vine run foo/foo.vi` would – if the file uses `mod <child>;` items – be an error, which would hopefully suggest you use `vine run foo` instead.)

Additionally, currently the name of a top-level source is inferred from the path. This would still be the case by default, but one could instead supply an alternate name by prefixing the argument with `<name>=`. For example, `vine run foo=/nix/store/0123456789abcdef-foo`. This, combined with the earlier change, fixes the issue with Nix.

I think this also integrates nicely with how I want to develop the build system. In particular, I would like it to continue to be the case that `vine run foo.vi` can run a Vine program with no dependencies, with no extra fluff. However, `vine run <name-of-package-in-workspace>` should also work as-is. This meshes nicely with this system, as the argument can *always* be a path to a source. In the `<name-of-package-in-workspace>` scenario, this is a path to the directory containing the package, which would be the source. And one could nicely run a package in a different directory with e.g. `vine run ../../other-project/foo`. In particular, the rules for top-level sources would just change to allow source directories to optionally have a package manifest file.

tjjfvi:
- resolves #498
- removes the ability to include a submodule from an explicit path (`mod foo = "./bar/baz.vi`)
- reworks the loader to perform all FS operations through an `FS` trait, allowing the use of virtual file systems (for e.g. a playground)

### DIFF
--- a/cli/src/vine_cli.rs
+++ b/cli/src/vine_cli.rs
@@ -555,7 +555,7 @@ impl TypedValueParser for ParseSource {
         Err(
           clap::Error::raw(
             ErrorKind::ValueValidation,
-            format!("invalid source name `{}`", name.to_string_lossy()),
+            format!("invalid source name `{}`", name.display()),
           )
           .with_cmd(cmd),
         )
@@ -568,7 +568,7 @@ impl TypedValueParser for ParseSource {
         Err(
           clap::Error::raw(
             ErrorKind::ValueValidation,
-            format!("could not detect name for source with path `{}`", path.to_string_lossy()),
+            format!("could not detect name for source with path `{}`", path.display()),
           )
           .with_cmd(cmd),
         )


cli/src/vine_cli.rs
```diff
@@ -503,3 +519,60 @@ fn colors(t: &impl IsTerminal) -> Colors {
...
+    if let Some(index) = bytes.iter().position(|&x| x == b'=') {
+      let name = unsafe { OsStr::from_encoded_bytes_unchecked(&bytes[0..index]) };
+      let path = unsafe { OsStr::from_encoded_bytes_unchecked(&bytes[index + 1..]) };
+      if let Some(name) = name.to_str().and_then(Ident::new) {
+        Ok(Source { name, path: path.to_owned().into() })
+      } else {
+        Err(
+          clap::Error::raw(
+            ErrorKind::ValueValidation,
+            format!("invalid source name `{}`", name.to_string_lossy()),
```
enricozb: Is the debug format of `&OsStr` not preferred?
```suggestion
            format!("invalid source name {name:?}"),
```

tjjfvi: I'm fine with it being lossy for the purpose of error messages, although I suppose `.display()` would probably at minimum be preferable.

vine/src/structures/diag.rs
```diff
@@ -73,12 +69,14 @@ macro_rules! diags {
...
-  AmbiguousImplicitMod { name: Ident }
-    ["ambiguous implicit submodule path; both `{name}.vi` and `{name}/{name}.vi` exist"]
+  CannotRead { path: String }
+    ["cannot read `{path}`"]
+  AmbiguousSubmodule { file_path: String, dir_path: String }
+    ["ambiguous submodule source; both `{file_path}` and `{dir_path}/` exist"]
+  MissingSubmodule { file_path: String, dir_path: String }
+    ["missing submodule; neither `{file_path}` nor `{dir_path}/` exist"]
+  InvalidSubmodule
+    ["cannot include a submodule from another file in a source file"]
```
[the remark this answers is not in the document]
enricozb: Ok this makes sense, thanks!

PR #504, opened 2026-02-02
vine: add `#[self_dual]`

### LINKED ISSUE #503: vine: allow structs and unions to be self-dual
The *dual* of a type constructor `Foo[A, B]` is the type constructor `~Foo[~A, ~B]`. A type constructor is said to be *self-dual* when it is equivalent to its dual, i.e. `Foo[A, B] = ~Foo[~A, ~B]` for all `A`, `B`.

This is equivalent to saying that the inverse distributes over the type constructor, i.e. `~Foo[A, B] = Foo[~A, ~B]`.

Currently, tuples and objects are self-dual, so `~(A, B) = (~A, ~B)` and `~{ a: A, b: B } = { a: ~A, b: ~B }`.

Semantically, this corresponds to the property that constructing a tuple is symmetric to destructing a tuple – both use the same node.

Currently, structs are never self-dual, so even a trivial struct like
```rs
struct Pair[A, B](A, B);
```
is not self-dual. There is no fundamental reason that this could not be self-dual.

It can be quite useful for a struct or [unsafe enum](https://github.com/VineLang/vine/issues/488) to be self-dual. Thus, add a `#[self_dual]` attribute which can be applied to a struct or unsafe enum, which makes the type self-dual.

For a struct or unsafe enum to be self-dual, its content type(s) must also be self-dual; for example,
```rs
#[self_dual] // ok
struct Foo[T](T, T);

#[self_dual] // error; ~(N32, T) != (N32, ~T)
struct Bar[T](N32, T);
```

One of the most significant uses for this is the `Array` type in root. It stores its values in a tree of type `Node[T]`. `Node` is an unsafe enum, and, with this feature, could be self-dual. This would allow for deduplicating a lot of the code in `Array`; for example, `pop_front` on a `&Node[T]` is equivalent to a `push_front` on `&Node[~T]`, but you can only turn a `&Node[T]` into a `&~Node[T]`, so to use this property, it must be the case that `~Node[T] = Node[~T]`.

tjjfvi:
- resolves #503 
- refactors `Array` to use a self-dual `Node`

### DIFF
--- a/vine/src/components/charter.rs
+++ b/vine/src/components/charter.rs
@@ -238,7 +238,7 @@ impl Charter<'_> {
             self.chart.unions[union_id].self_dual = true;
           }
           _ => {
-            self.diags.error(Diag::BadTestAttr { span });
+            self.diags.error(Diag::BadSelfDualAttr { span });
           }
         },
       }


vine/src/structures/types.rs
```diff
@@ -244,12 +252,16 @@ impl Types {
       (TypeKind::IfConst(a, t, e), TypeKind::IfConst(b, u, f)) if a == b => {
         self.unify(*t, u.invert_if(inverted)).and(self.unify(*e, f.invert_if(inverted)))
       }
+      (TypeKind::Struct(StructId(i), self_dual, a), TypeKind::Struct(StructId(j), _, b))
+      | (TypeKind::Union(UnionId(i), self_dual, a), TypeKind::Union(UnionId(j), _, b))
+        if i == j && (*self_dual || !inverted.0) =>
```
nilscrm: Is the `self_dual` check here enough. I would imagine it would make a difference if the parameters to a generic struct are self-dual or not. I.e. a struct is probably only self-dual if all its generics are self-dual.

vine/src/features/struct_.rs
```diff
@@ -129,6 +130,9 @@ impl Resolver<'_> {
       }
     );
     self.annotations.record_signature(struct_def.span, hover);
+    if struct_def.self_dual && !self.types._self_inverse(data, true) {
+      self.diags.error(Diag::NotSelfDual { span: struct_def.span });
```
nilscrm: Do we also want to provide a warning if a type without generics is marked as self-dual?

tjjfvi: Self-duality of a type with no generics is equivalent to being self-inverse. A type that consists of a self-dual type constructor all of whose arguments are self-inverse is itself self-inverse, and this is vacuously satisfied when there are no type parameters.

nilscrm (approved):

PR #511, opened 2026-02-08
vine: add configurable constants

### LINKED ISSUE #479: vine: configurable constants

tjjfvi:
- resolves #479 

Based on #494

### DIFF
--- a/vine/src/components/charter.rs
+++ b/vine/src/components/charter.rs
@@ -252,7 +252,7 @@ impl Charter<'_> {
         },
         AttrKind::Configurable => {
           let Some(concrete_const_id) = concrete_const_id else {
-            self.diags.error(Diag::BadTestAttr { span });
+            self.diags.error(Diag::BadConfigurableAttr { span });
             continue;
           };
 


vine/src/components/charter.rs
```diff
@@ -246,6 +250,14 @@ impl Charter<'_> {
             self.diags.error(Diag::BadSelfDualAttr { span });
           }
         },
+        AttrKind::Configurable => {
+          let Some(concrete_const_id) = concrete_const_id else {
+            self.diags.error(Diag::BadTestAttr { span });
```
nilscrm:
```suggestion
            self.diags.error(Diag::BadConfigurableAttr { span });
```

PR #512, opened 2026-02-08
vine: add `#[impl]` attribute

### LINKED ISSUE #330: vine: impl derivation shorthand
Add a shorthand for deriving implementations of several one-type-parameter traits, such that
```rs
#[impl(fork: Fork, drop: Drop, show: Show)]
pub struct Foo[A, B, C](...);
```

would be equivalent to
```rs
pub struct Foo[A, B, C](...);

pub mod Foo {
  pub impl fork[A, B, C; Fork[A], Fork[B], Fork[C]]: Fork[Foo[A, B, C]];
  pub impl drop[A, B, C; Drop[A], Drop[B], Drop[C]]: Drop[Foo[A, B, C]];
  pub impl show[A, B, C; Show[A], Show[B], Show[C]]: Show[Foo[A, B, C]];
}
```

This would reduce boilerplate for common cases by quite a bit.

Once #210 is implemented, this could be reduced even further:
```rs
#[impl(Fork, Drop, Show)]
pub struct Foo[A, B, C](...);
```

tjjfvi:
- resolves #330

### DIFF
--- a/lsp/tree-sitter-vine/grammar.js
+++ b/lsp/tree-sitter-vine/grammar.js
@@ -132,6 +132,7 @@ module.exports = grammar({
       "unsafe",
       "safe",
     ],
+    none: _ => [],
   },
 
   rules: {
@@ -330,7 +331,8 @@ module.exports = grammar({
       seq(
         "#[",
         optional("safe"),
-        $.ident,
+        // @ts-ignore
+        reserved("none", $.ident),
         optional(choice(seq("=", $.string), seq("(", $._expr, ")"))),
         "]",
       ),


vine/src/features/impl_.rs
```diff
@@ -314,6 +327,22 @@ impl Resolver<'_> {
...
         self.chart.builtins.drop.ok_or(Diag::MissingBuiltin { span: *span, builtin: "Drop" })
       }
+      DeriveKind::Trait(trait_) => {
+        let span = trait_.span;
+        if trait_.generics.is_some() {
+          return Err(Diag::DerivePassedArguments { span });
+        }
+        match self.resolve_path(self.cur_def, trait_, "trait", |d| d.trait_kind) {
+          Ok(DefTraitKind::Trait(trait_id)) => {
+            if self.sigs.type_params[self.chart.traits[trait_id].generics].params.len() == 1 {
```
nilscrm: How do we want to handle derives for traits with not exactly one type parameter? I'm thinking e.g. all the binary arithmetic and bitwise operations.

Issue #460, opened 2025-12-16
ivm: generalized branches


tjjfvi:
Currently, branch nodes have the form
```
value = ?(if_zero if_nonzero ctx)
```
and have the following reduction rules:
```
0 = ?(if_zero if_nonzero ctx)
-----------------------------
if_zero = ctx
if_nonzero = _

[n+1] = ?(if_zero if_nonzero ctx)
-----------------------------
if_zero = _
if_nonzero = ctx
```

This works well for branching on booleans. However, it would be desirable to have higher-arity branches. For example, if you have an enum with 4 variants, and want to match on it, you would currently have to write something like follows:
```
value = dup(v @n32_lt(2 ?(::if_lt_two ::if_ge_two x(v ctx))))

::if_lt_two {
  x(value ctx)
  value = ?(::if_zero ::if_one ctx)
}

::if_ge_two {
  x(value ctx)
  value = @n32_eq(2 ?(::if_three ::if_two ctx))
}

::if_zero { ... }
::if_one { ... }
::if_two { ... }
::if_three { ... }
```
This involves a lot of interactions. Currently, thus, Vine does not use N32s to represent enums, instead using [Scott encoding](https://en.wikipedia.org/wiki/Mogensen%E2%80%93Scott_encoding). However, though this is more efficient to match on, it is much less space-efficient, especially with large enums.

One could imagine introducing higher-arity branch nodes, like `?(if_zero if_one if_two if_three ctx)`, but this would be difficult to implement efficiently. This is because, though the current branch nodes are logically ternary, the IVM only has binary nodes. So under the hood, `?(a b c)` is represented as `?(?(a b) c)` with the following reduction rules:
```
0 = ?(branches ctx)
-----------------------------
branches = ?(ctx _)

[n+1] = ?(branches ctx)
-----------------------------
branches = ?(_ ctx)
```
A higher-arity branch node would thus be represented as something like `?(?(?(a b) ?(c d)) e) and implemented as something like
```
0 = ?(branches ctx)
-----------------------------
branches = ?(?(ctx _) _)

1 = ?(branches ctx)
-----------------------------
branches = ?(?(_ ctx) _)

2 = ?(branches ctx)
-----------------------------
branches = ?(_ ?(ctx _))

3 = ?(branches ctx)
-----------------------------
branches = ?(_ ?(_ ctx))
```
These complex reduction rules would be difficult to implement, especially in hardware.

[...]

PR #514, opened 2026-02-10
ivy: generalized branches

enricozb:
resolves: #460

Branch nodes are now represented as extrinsic function calls. Specifically, `?(b_0 .. b_i c)` becomes `@branch(?[b_0 .. b_i] c)`, where `?[..]` is a lookup extrinsic value. This thus removes the need for the `Tag::Branch` variant.

Lookup extrinsic values can be directly created with the syntax `?[a b .. n]`, where `a b .. n` are all nilary and trivially-copyable nodes (lookup nodes also satisfy this).

### DIFF
--- a/tests/snaps/aoc_2024/day_01/stats
+++ b/tests/snaps/aoc_2024/day_01/stats
@@ -1,20 +1,20 @@
 
 Interactions
-  Total               14_608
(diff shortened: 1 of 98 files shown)


tjjfvi:
integrated into #545

PR #518, opened 2026-02-12
lsp: fix bug when compiling syntax errors

tjjfvi:
Fixes a bug that after compiling with syntax errors, the loaded modules would not be cleared, meaning that the next compilation would have all files duplicated, causing a ton of problems.

@nilscrm can you confirm this fixes the issue you encountered?


nilscrm (approved):
Thanks as far as I can see this fixes the problem.

PR #519, opened 2026-02-13
root: add float operations

nilscrm:
* Add generic `Float` trati for `F32`/`F64`
* Implement `parse` and `to_string` for `Float`
* Add `F64::exp`
* Add `F64::ln`
* Add extrinsics for `{F32,F64}::sqrt`


root/numeric/F64/F64.vi
```diff
@@ -1,10 +1,15 @@
 
 use ops::{Cast, arithmetic::{Add, Div, Mul, Neg, Pos, Pow, Rem, Sub}, comparison::{Eq, Le, Lt}};
 
+pub mod common;
```
tjjfvi: I think this should be private

PR #526, opened 2026-02-19
root: refactor List and increase parallelism

nilscrm:
Previously a lot of List method implementations prevented a lot of parallelism. Operations like `new`, `from_fn`, `reversed` build their entire output list before returning list length, as they would modify the length in each loop iteration with a push. Operations like `map` would not have this problem as it handled the start/end explicitly which didn't involve length altering calculation. This PR makes two changes to the list implementation.

* `IterRef` is not able to modify the tail anymore. This is a rare use case and not allowing it allows the length of the list to propagate immediately instead of requiring the iterator to be dropped first. 
* The circular list structure is now implemented with a `Stream` type. This allows convenient methods on it like push and pop, making the code more readable while not doing unnecessary length calculations and allowing parallelism

The PR also makes various refactors using modern Vine features.

### DIFF
--- a/tests/snaps/aoc_2024/day_01/stats
+++ b/tests/snaps/aoc_2024/day_01/stats
@@ -1,20 +1,20 @@
 
 Interactions
-  Total               14_608
-  Annihilate           7_874
+  Total               13_629
+  Annihilate           6_935
   Commute                129
-  Copy                 1_534
-  Erase                1_818
-  Expand               1_160
-  Call                 1_382
+  Copy                 1_459
+  Erase                1_773
+  Expand               1_156
+  Call                 1_466
   Branch                 711
 
 Parallelism
-  Depth                3_773
-  Breadth                  3
-  Speedup                 59 cB
+  Depth                3_391
+  Breadth                  4
+  Speedup                 60 cB
 
 Memory
-  Heap                18_672 B
-  Allocated          318_752 B
-  Freed              318_752 B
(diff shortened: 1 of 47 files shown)


root/data/List.vi
```diff
@@ -121,7 +120,7 @@ pub mod List {
   }
 
   pub fn .split_at[T](self: List[T], i: N32) -> (List[T], List[T]) {
-    if i >= self.len() {
+    assert i < self.len() else {
       return (self, []);
     }
```
enricozb:
```suggestion
    assert i < self.len() else { (self, []) }
```

root/data/List.vi
```diff
@@ -427,19 +409,19 @@ pub mod List {
...
   pub impl iter_ref[T]: Cast[&List[T], IterRef[T]] {
-    fn cast(&List[T](len, buf, end)) -> IterRef[T] {
-      IterRef(len, &buf, &(len, end))
+    fn cast(&List[T](len, Stream(buf, _))) -> IterRef[T] {
+      IterRef(len, &buf)
     }
   }
 
-  pub struct IterRef[T]((N32, &Buf[T], &(N32, ~Buf[T])));
+  pub struct IterRef[T]((N32, &Buf[T]));
```
enricozb:
```suggestion
  pub struct IterRef[T](N32, &Buf[T]);
```

tjjfvi (approved):

PR #529, opened 2026-02-20
vine: report fork/drop errors in analyzer

### LINKED ISSUE #359: vine: UninitializedVariable span at read
Currently the `variable of type `{ty}` read whilst uninitialized` compiler error is marked at the definition of the local variable not when the read happens. This makes debugging a lot harder.

Example:
```rust
pub fn .dpll(&formula: &Formula, assignment: Model) -> Option[Model] {
  loop {
    if formula.is_empty() {
      return Some(assignment);
    }

    if formula.has_empty_clause() {
      return None;
    }

    if formula.find_unit() is Some(literal) {
      assignment.insert(literal.variable(), literal.is_positive());
      let Some(formula) = formula.assign(literal) else { return None; };
      continue;
    }

    if formula.find_pure() is Some(literal) {
      assignment.insert(literal.variable(), literal.is_positive());
      let Some(formula) = formula.assign(literal) else { return None; };
      continue;
    }

    break;
  }

  let Some(branch_literal) = formula.choose_literal() else {
    return Some(assignment);
  };

  if formula.assign(branch_literal) is Some(next_formula) {
    let new_assignment = assignment;
    new_assignment.insert(branch_literal.variable(), true);
    if next_formula.dpll(new_assignment) is Some(solution) {
      return Some(solution);
    }
  }

  if formula.assign(branch_literal.negated()) is Some(next_formula) {
    assignment.insert(branch_literal.variable(), false);
    if next_formula.dpll(assignment) is Some(solution) {
      return Some(solution);
    }
  }

  None
}
```
where the parameter `assignment` is marked as read whilst uninitialized.

tjjfvi:
Moves the reporting of fork/drop errors from the emitter to the analyzer, meaning that the emitter does not need to be run when checking without building.

Also fixes #359 – now fork/drop errors are (mostly) reported at the offending invocation of the variable, rather than always on its declaration. Additionally, errors when something cannot be forked are consistently "cannot fork" rather than "read whilst uninitialized", and (sometimes) provide a warning of where the variable was used for a second time.

There are still some edge cases where it cannot find the offending invocation; in these cases it falls back to reporting the issue at the declaration. Additionally, the warning can only be issued when the second invocation is in the same stage, and will not be reported otherwise.

There are some broader changes I want to make to the analyzer in the near future, but it's good to get this in at a minimum.

### DIFF
--- a/vine/src/components/analyzer.rs
+++ b/vine/src/components/analyzer.rs
@@ -331,7 +331,7 @@ impl Analyzer<'_> {
         Step::Transfer(transfer) => {
           let (i, b, o) =
             self.interfaces[transfer.interface].wires.get(&local).copied().unwrap_or_default();
-          let (i, b, o) = if info.inv.0 { (o, b, i) } else { (i, b, o) };
+          let (i, o) = if info.inv.0 { (o, i) } else { (i, o) };
           (i, b, o, None)
         }
         _ => (false, false, false, None),


vine/src/components/analyzer.rs
```diff
@@ -100,9 +105,12 @@ impl Analyzer<'_> {
     let mut forwards = Vec::new();
     let mut backwards = Vec::new();
 
+    let mut strict_non_fork = false;
```
nilscrm: What does this variable mean? Contrary to the name it looks like the variable can be forked when `local.fork.is_some()` even if `strict_non_fork` is true.

tjjfvi: `strict_non_fork` means that non-forkable types have to be handled strictly such that every read is treated as a read-barrier (a move, in the Rust sense). This is only needed in the repl, as it will only not read the value if it is uninitialized, so the variables needs to be considered uninitialized if the variable has been moved. In other contexts, it's preferable to not treat non-forkable reads as read-barriers, because then the error will be "cannot fork" instead of "read whilst uninitialized".

This is a bit of a hack/patch; ideally when I rework the analyzer's effect system this would no longer be necessary.

PR #534, opened 2026-02-23
vine: `vine test` cli command improvements

enricozb:
`vine test <source>` runs all `#[test]` fns in `source` and prints `ok` and `FAILED` accordingly. Standard output is captured and printed only in the case of failure, unless `--no-capture` is set.

resolves: #118

### DIFF
--- a/cli/src/common.rs
+++ b/cli/src/common.rs
@@ -42,7 +42,7 @@ pub struct RunArgs {
   breadth_first: bool,
   #[arg(long, short = 'H', value_parser = parse_size)]
   heap: Option<usize>,
-  pub argv: Vec<String>,
+  pub args: Vec<String>,
 }
 
 impl RunArgs {
@@ -87,7 +87,7 @@ impl RunArgs {
     let mut extrinsics = Extrinsics::default();
 
     host.register_default_extrinsics(&mut extrinsics);
-    host.register_runtime_extrinsics(&mut extrinsics, &self.argv, io_input_fn, io_output_fn);
+    host.register_runtime_extrinsics(&mut extrinsics, &self.args, io_input_fn, io_output_fn);
     host.insert_nets(nets);
 
     let main = host.get("::").expect("missing main");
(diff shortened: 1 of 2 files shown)


cli/src/vine_cli.rs
```diff
@@ -185,44 +185,74 @@ pub struct VineTestCommand {
...
-      eprintln!("test path {test_path:?} does not exist");
-      exit(1);
-    };
+    let mut failed = false;
+    for (path, test_id) in tests {
+      compiler.insert_main_net(&mut nets, test_id);
 
-    let main_fragment_id = compiler.resolutions.fns[*test_concrete_fn_id];
-    compiler.insert_main_net(&mut nets, main_fragment_id);
+      eprint!("test {path} ... ");
```
tjjfvi: This header is pretty visually indistinct from the output; maybe:
```suggestion
      eprint!("{grey}test{reset} {bold}{path}{reset} ... ");
```

tests/tests.nix
```diff
@@ -267,12 +267,14 @@ let
       let
         vi = just ./misc/basic_test.vi;
         result = pkgs.runCommand "misc-basic_test" { } ''
```
tjjfvi: I think this is a great use-case for the `interactive` utility I have in the tests.

I think the `"#basic_test::test"` can be removed, but can you add a test that filters to e.g. just run `test_proper_div` and confirm it succeeds?

ivy/src/optimize/prune.rs
```diff
@@ -3,13 +3,15 @@ use std::collections::HashSet;
...
-    nets.retain(|name, _| keep.contains(name));
+pub fn prune(nets: &mut Nets, entrypoints: &[String]) {
+  let mut prune = Prune { nets, keep: HashSet::new() };
+  for global in entrypoints.iter().map(String::as_str).chain(["::"]) {
+    if nets.contains_key(global) {
+      prune.visit_global(global);
+    }
   }
+  let keep = prune.keep;
+  nets.retain(|name, _| keep.contains(name));
```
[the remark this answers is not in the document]
enricozb: I didn't realize the utility here for graceful behavior in the event of a misbehaving compiler

PR #543, opened 2026-03-03
vine: implicit `else { _ }` in `if`/`for`/`while`

### LINKED ISSUE #532: vine: evaluate to `Default::default` when missing `else`
For `if foo { bar }`, instead of ignoring `bar` and evaluating to `()`, have it be equivalent to `if foo { bar } else { Default::default }`. Then things like `if foo { Some(bar) } else { None() }` could be written more simply `if foo { Some(bar) }`.

Similarly for `while` and `for`. Not sure about `when`.

I think this is better than #344 as it is more general, more explicit, and avoids the issue @nilscrm describes.

enricozb:
resolves #532

my main issue with the implementation is the loss of the `MissingElse` diagnostics.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -4,6 +4,7 @@
   "words": [
     "antinodes",
     "brainfuck",
+    "carnegieplein",
     "centimanes",
     "cmps",
     "codepoint",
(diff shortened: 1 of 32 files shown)


tjjfvi (approved):
Looks good. I rebased this onto #545 and pushed a few tweaks.

PR #545, opened 2026-03-08
ivy: hedera

### LINKED ISSUE #54: ivm: rework ext system

### LINKED ISSUE #159: ivy: rework label set calculation

### LINKED ISSUE #257: ivy: agnosticism
Currently, the interaction system of Ivy has been designed so as to be convenient for implementing in IVM. (Though I have previously referred to this system as Ivy's interaction system, I will refer to it as IVM's interaction system within this issue for clarity.)

Vine also has an interaction system, internal to its semantics, that does not align with the current Ivy interaction system. Currently, the Vine compiler converts nets in its interaction system to IVM's interaction system during emission.

However, what's convenient for implementing in IVM is not necessarily what is convenient for other environments. Thus, some hypothetical other environment (let's say, TC, for Temporary Codename) will have a different interaction system, TC's interaction system. To run Vine code in TC, one would then need to translate the IVM-system nets outputted by the Vine compiler to the TC system. However, the IVM system is quite general, and might be difficult to translate faithfully – particularly, the semantics around global nets and extrinsics could be difficult to match.

Much more convenient than translating IVM-system nets to TC-system nets would be translating Vine-system nets to TC-system nets. Currently, though, there is no good way to do this, as Vine-system nets are never exposed by the compiler.

Therefore, to support more environments, the Vine compiler should output Vine-system nets, rather than IVM-system nets. Thus, one needs a textual syntax for Vine-system nets. Furthermore, one still wants to have a textual syntax for IVM-system nets (for example, to capture the result of pre-reducing), or even TC-system nets.

Thus, Ivy will become a general textual syntax for nets, agnostic of the interaction system. A net in any interaction system can be written in Ivy, so one can have Vine-system Ivy, IVM-system Ivy, or TC-system Ivy. 
Agent symbols from a particular interaction system would be namespaced. This would even allow nets in a mixture of interaction systems; you might have a net using `vine:ref` nodes (from the Vine system), alongside `ivm:io_print_char` nodes (from the IVM system).

Then, tools can be developed to convert Ivy nets between different interaction systems. The current Vine-to-IVM transformation implemented in the Vine compiler will become part of the IVM project, and convert Vine-system Ivy to IVM-system Ivy. 

This greatly improves the story for TC – the Vine compiler will output nets in Vine-system Ivy, and so all TC has to do is translate Vine-system Ivy to TC-system Ivy.

### LINKED ISSUE #370: ivy: check & error on unbound variables
Currently unbound variables will only trigger an error in debug mode; in release mode, they'll be carried all the way down into unpaired registers in the IVM which will cause wires to be created between different global expansions in very cursed and wrong ways.

### LINKED ISSUE #461: vine: numerically encoded enums
Currently, enums are [Scott encoded](https://en.wikipedia.org/wiki/Mogensen%E2%80%93Scott_encoding), so something like
```rs
enum Foo {
  A(N32),
  B(F32),
  C(),
}

let a = Foo::A(4);
let b = Foo::B(6.0);
let c = Foo::C();

match foo {
  Foo::A(a) { ... }
  Foo::B(b) { ... }
  Foo::C() { ... }
}
```
is represented like
```rs
a = enum(enum(4 x) enum(_ enum(_ x)))
b = enum(_ enum(enum(6.0 x) enum(_ x)))
c = enum(_ enum(_ enum(x x)))

foo = enum(::match_a enum(::match_b enum(::match_c ctx)))
```
This is efficient for matching, but very inefficient in terms of space. Once #460 is implemented, change enums to be represented with a numeric tag, like so:
```
a = enum(0 4)
b = enum(1 6.0)
c = enum(2 _)

foo = enum(?(?^(?^(::match_a ::match_c) ::match_b) x(ctx data)) data)
```

This is similarly efficient for matching, but much, much more space-efficient.

### LINKED ISSUE #482: vine: rework inline ivy syntax
The current inline ivy syntax is quite messy and outdated. It uses an `inline_ivy!` token, which is quite cursed.

I think ultimately it would be nice for this to be a compiler-builtin macro (#301). Thus, rework the syntax for now to look more like the intended macro syntax.

Furthermore, the current syntax binds Ivy variables to Vine values/spaces with `x <- value` and `x -> space`. This usually ends up being very verbose, is a bit of an abuse of the `->` token, and is the only use of the `<-` token.

Thus, change the syntax for inline ivy:
```
// old
inline_ivy! (io0 <- io, io1 -> io) -> Char {
  char
  io0 = @io_read_char(char io1)
}

// new
$ivy(
  char
  ${io} = @io_read_char(char ${~io})
)
```
Particularly, the Ivy variable binding section is removed, and replaced with the ability to embed Vine expressions inside the Ivy with `${...}`. Alternate syntax for value/space binding is unnecessary, as spaces can just be inverted into values before embedding.

Furthermore, this removes the `inline_ivy!` token. `$ivy` will not be its own token, but rather the token `$` followed by the identifier `ivy`, which the parser will recognize in this context (similarly to how attributes are handled).

tjjfvi:
- resolves #544 
- resolves #257 
- resolves #370 
- resolves #482 
- resolves #159 
- resolves #54 
- resolves #461


ivy/src/net.rs
```diff
@@ -0,0 +1,391 @@
...
+
+#[derive(Debug, Clone)]
+pub struct FlatNode {
+  pub name: Name,
+  pub pri: Wire,
+  pub aux: Vec<Wire>,
+}
+
+#[allow(non_snake_case)]
+pub fn FlatNode(name: impl Into<Name>, pri: Wire, aux: impl IntoIterator<Item = Wire>) -> FlatNode {
```
nilscrm: Since `FlatNode` is not used that often, any reason for not using the more common `FlatNode::new`

ivy/src/prune.rs
```diff
@@ -0,0 +1,53 @@
...
+  net::FlatNet,
+};
+
+pub fn prune(
+  table: &Table,
+  nets: &mut HashMap<NameId, FlatNet>,
+  entry: impl IntoIterator<Item = NameId>,
+) {
+  let mut seen = HashSet::new();
+  walk(table, nets, entry, |name| seen.insert(name));
```
nilscrm:
```suggestion
  for name in entry {
    visit(table, nets, name, &mut |name| seen.insert(name));
  }
```

ivy/src/prune.rs
```diff
@@ -0,0 +1,53 @@
...
+          } else {
+            for &child in &node.name.children {
+              visit(table, nets, child, f);
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
nilscrm:
```suggestion
fn visit(
  table: &Table,
  nets: &HashMap<NameId, FlatNet>,
  name: NameId,
  f: &mut impl FnMut(NameId) -> bool,
) {
  if f(name) {
    for &child in &table.name(name).children {
      visit(table, nets, child, f);
    }
    if let Some(net) = nets.get(&name) {
      for node in &net.nodes {
        if let Some(name) = table.has_name(&node.name) {
          visit(table, nets, name, f);
        } else {
          for &child in &node.name.children {
            visit(table, nets, child, f);
          }
        }
      }
    }
  }
}
```

ivy/src/translate.rs
```diff
@@ -0,0 +1,92 @@
...
+    table: &mut Table,
+    net: &mut FlatNet,
+    final_nodes: &mut Vec<FlatNode>,
+    base: RuleId,
+  ) {
+    let node = net.nodes.pop().unwrap();
+    if let Some(rules) = self.rules.get(&node.name.path)
+      && let Some((id, rule)) = rules.iter().find(|&&(id, _)| id >= base)
+    {
+      let base = net.nodes.len();
```
nilscrm: It's a little confusing if `base` is both the base rule and the originally number of nodes

ivy/src/engine.rs
```diff
@@ -0,0 +1,631 @@
...
+    (node, (0..len).map(move |i| Port(PortRef { node, index: PortIndex(i) })))
+  }
+
+  pub fn remove_node(&mut self, node_id: NodeId) -> impl use<> + ExactSizeIterator<Item = Port> {
+    let node = &mut self.nodes[node_id];
+    let net_id = node.net;
+    let net = &mut self.nets[net_id];
+    self.dead_nodes.push(node_id);
+    let ports = node.ports.keys().map(move |index| Port(PortRef { node: node_id, index }));
+    if net.not_nodes * 2 >= net.maybe_nodes.capacity() {
```
nilscrm: As far as I can see the only place than can make this condition true is `finish_edit` so that might be a better place to do the tightening.

ivy/src/engine.rs
```diff
@@ -0,0 +1,631 @@
...
+      self.tighten_net(net_id);
+    }
+    if self.not_dirty * 2 >= self.maybe_dirty.capacity() {
```
nilscrm: Same with this condition

ivy/src/engine.rs
```diff
@@ -0,0 +1,631 @@
...
+    }
+    let net_id = self.nodes[node].net;
+    net.awaiting.push((node, net_id));
+    self.nets[net_id].pending += 1;
+    false
+  }
+
+  pub fn graft_net_n<const N: usize>(&mut self, from_id: NetId, into_id: NetId) -> [Port; N] {
+    exact_size(self.graft_net(from_id, into_id))
+  }
```
nilscrm, line 374:
```suggestion
```

ivy/src/engine.rs
```diff
@@ -0,0 +1,631 @@
...
+      };
+      let new_id = self.nodes.insert(new);
+      into.maybe_nodes.push(new_id);
+      if dirty {
+        self.maybe_dirty.push(new_id);
+      }
+      self.nodes[old_id].net.0 = new_id.0; // crimes
+    }
+    into.pending += from.pending;
+    into.pending += from.pending;
```
nilscrm: Why is this done twice?

ivy/src/engine.rs
```diff
@@ -0,0 +1,631 @@
...
+{
+  fn register(self, rules: &mut Rules<'a>) {
+    let Rewrite(paths, f) = self;
+    rules.register_rewrite(paths, f);
+  }
+}
+
+pub trait Reducer {
+  fn visit(&mut self, engine: &mut Engine, net: NetId, node: NodeId);
+}
```
nilscrm:
```suggestion
```

ivy/src/engine/common.rs
```diff
@@ -0,0 +1,237 @@
...
+      return false;
+    }
+    let a = engine.remove_node(a);
+    let b = engine.remove_node(b);
+    engine.link_all(a, b);
+    true
+  })
+}
+
+pub fn erase<'r>(
```
nilscrm, line 44: Thinking about it, an erase is simply a commute between a nilary and an n-ary node. I don't know if we actually want to make use of this fact, but I think it's interesting.

ivy/src/engine/common.rs
```diff
@@ -0,0 +1,237 @@
...
+      }
+    }
+    true
+  })
+}
+
+pub fn eta_reduce<'r>(
+  combinators: impl IntoIterator<Item = PathId>,
+  values: impl IntoIterator<Item = PathId>,
+  nils: impl IntoIterator<Item = PathId> + Clone,
```
nilscrm, line 97: Look mom, it's me :)

ivy/src/engine/common.rs
```diff
@@ -0,0 +1,237 @@
...
+      for port in ports {
+        let [port_] = engine.remove_node_n(engine.follow(*port).node);
+        engine.remove_wire(port, port_);
+      }
+      let [pri_] = engine.insert_node_n(net, name);
+      engine.link(pri, pri_);
+      true
+    } else {
+      let (a, a_aux) = (node, aux);
+      let b = aux_.clone().find(|x| !x.is_principal()).unwrap().node;
```
nilscrm, line 129:
```suggestion
      let b = aux_.find(|x| !x.is_principal()).unwrap().node;
```

ivm/src/runtime/ext.rs
```diff
@@ -111,8 +91,7 @@ impl<'ivm> Extrinsics<'ivm> {
 /// - the next 45 bits are the payload.
 /// - the last 3 bits are `Tag::ExtVal as u16` for compatibility with [`Port`].
 ///
-/// [`Port`]: crate::ivm::Port
-#[derive(Clone, Copy)]
+/// [`Port`]: crate::vm::Port
```
nilscrm:
```suggestion
/// [`Port`]: crate::runtime::Port
```

ivm/src/runtime/ext.rs
```diff
@@ -401,9 +368,9 @@ impl<'ivm> ExtTyId<'ivm> {
 /// The lowest 14 bits denote an index into one of `ext_split_fns` or
 /// `ext_merge_fns`.
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct ExtFn<'ivm>(u16, PhantomData<fn(&'ivm ()) -> &'ivm ()>);
+pub struct ExtFnId<'ivm>(u16, PhantomData<fn(&'ivm ()) -> &'ivm ()>);
```
nilscrm, line 371: This is not part of the PR but since I see it right now I thought I'd mention it. Does an split ext fn with swapped arguments make sense? Otherwise I think it would be nicer to have the highest bit indicate whether the function is merge of split and in the case it is merge, have the second-highest bit denote whether the arguments are swapped.

ivm/src/host/ext.rs
```diff
@@ -0,0 +1,476 @@
...
+  pub fn register_ext_fn<const I: usize, const O: usize, W>(
+    &mut self,
+    table: &mut Table,
+    path: &str,
+    f: impl ExtFn<'ivm, I, O, W>,
+  ) {
+    let name = table.add_path_name(path);
+    let f = f.register(self, table);
+    let pair = self.register_ext_ty();
+    let ref_ = self.register_ext_ty();
```
nilscrm: I feel like being a little more explicit here is nice for documentation purposes and to not overly rely on the type inference. The same holds for other invocations of `register_ext_ty`.
```suggestion
    let pair = self.register_ext_ty::<Pair>();
    let ref_ = self.register_ext_ty::<Ref>();
```

ivm/src/host/ext.rs
```diff
@@ -0,0 +1,476 @@
...
+  impl<'r, 'p, 'ivm, const I: usize, const O: usize, W, F: ExtFn<'ivm, I, O, W>>
+    Register<HostTable<'ivm, 'r>> for _ExtFn<'p, I, O, W, F>
+  {
+    fn register(self, registry: &mut HostTable<'ivm, 'r>) {
+      registry.host.register_ext_fn(registry.table, self.path, self.f);
+    }
+  }
+
+  _ExtFn { path, f, phantom: PhantomData }
+}
```
nilscrm, line 158: This is somewhat cursed. Is it possible to just have a normal struct, a `new` function and implement `ExtFn` for it?

ivm/src/host/ext.rs
```diff
@@ -0,0 +1,476 @@
...
+  ) -> impl use<'ivm, N, T, W> + Live<'ivm> + Fn(&mut Rt<'ivm, '_>, Self, [Wire<'ivm>; N]) {
+    let handle_outputs = T::register(host, table);
+    move |rt, result, wires| {
+      let Ok(outputs) = result else { return error(rt, wires) };
+      handle_outputs(rt, outputs, wires);
+    }
+  }
+}
+
+pub trait ExtFn<'ivm, const I: usize, const O: usize, W> {
```
nilscrm, line 272: I'm kind of torn on this API. On the one hand it's quite magical that you can just provide different kind of functions to register ext fns but on the other hand it's also quite magical in the sense that you just have to know what things implement `ExtFn`. The other option of just having register methods on the host would allow one to have the lsp give options for the different ways of registering an ext fn (e.g. methods `register`, `register_with_runtime`, `register_with_host`, etc.) and give type hints for the arguments for each.

ivm/src/host/ext.rs
```diff
@@ -0,0 +1,476 @@
...
+}
+
+impl<'ivm, T: ExtTyBoxed<'ivm>> ExtTyRegister<'ivm> for Boxed<T> {
+  type With<'x> = Boxed<T::With<'x>>;
+}
+
+#[derive(Default, Debug, Clone, Copy)]
+pub struct Invalid;
+
+pub trait ExtInput<'ivm, W>: Sized {
```
nilscrm, line 178: I really like the way inputs and outputs are derived

util/src/arithmetic.rs
```diff
@@ -0,0 +1,176 @@
...
+  R: Define<f64, (), W9>,
+  R: Define<(f64, f64), f64, W10>,
+  R: Define<f64, f64, W11>,
+  R: Define<(f64, f64), bool, W12>,
+  R: Define<(u32, u32), f64, W13>,
+  R: Define<f64, (u32, u32), W14>,
+  R: Define<f64, u32, W15>,
+  R: Define<u32, f64, W16>,
+  R: Define<f32, f64, W17>,
+  R: Define<f64, f32, W18>,
```
nilscrm, line 71:
```suggestion
pub fn arithmetic<R, E: 'static + Send + Sync + Copy, W21, W11, W12, W10, W21b, W21r>(
  err: E,
) -> impl Register<R>
where
  R: Define<(u32, u32), u32, W21>,
  R: Define<(u32, u32), bool, W21b>,
  R: Define<(u32, u32), Result<u32, E>, W21r>,
  R: Define<(f32, f32), f32, W21>,
  R: Define<f32, f32, W11>,
  R: Define<(f32, f32), bool, W21b>,
  R: Define<u32, f32, W11>,
  R: Define<f32, u32, W11>,
  R: Define<f64, (f64, f64), W12>,
  R: Define<f64, (), W10>,
  R: Define<(f64, f64), f64, W21>,
  R: Define<f64, f64, W11>,
  R: Define<(f64, f64), bool, W21b>,
  R: Define<(u32, u32), f64, W21>,
  R: Define<f64, (u32, u32), W12>,
  R: Define<f64, u32, W11>,
  R: Define<u32, f64, W11>,
  R: Define<f32, f64, W11>,
  R: Define<f64, f32, W11>,
```

ivm/src/host/ext/common.rs
```diff
@@ -0,0 +1,565 @@
...
+  }
+
+  #[inline(always)]
+  unsafe fn from_payload_static(payload: Word) -> u32 {
+    (payload.bits() >> 3) as u32
+  }
+}
+
+impl<'ivm> ExtTyCast<'ivm> for u64 {
+  const COPY: bool = true;
```
nilscrm: Should this be `false`?

ivm/src/host/ext/common.rs
```diff
@@ -0,0 +1,565 @@
...
+    f32::from_bits(unsafe { u32::from_payload_static(payload) })
+  }
+}
+
+impl<'ivm> ExtTyRegister<'ivm> for f64 {
+  type With<'x> = Self;
+}
+
+impl<'ivm> ExtTyCast<'ivm> for f64 {
+  const COPY: bool = true;
```
nilscrm: should this be `false`?

ivm/src/runtime/interact.rs
```diff
@@ -88,15 +88,10 @@ impl<'ivm, 'ext> IVM<'ivm, 'ext> {
     use Tag::*;
     match ((a.tag(), a), (b.tag(), b)) {
       sym!((Wire, _), _) | sym!((Erase | ExtVal, _)) => unreachable!(),
```
nilscrm: This unreachable case should probably also include the `(Graft, Erase)` case.

ivy/src/name.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    impl $crate::name::FromTable<'_> for $name {
+      fn build(table: &mut $crate::name::Table) -> Self {
+        $(let $key = guide!(impl table $type $str);)*
+        Self { $($key),* }
+      }
+    }
+  };
+
+  (type ) => { $crate::name::PathId };
```
nilscrm:
```suggestion
```

ivm/src/program.rs
```diff
@@ -0,0 +1,281 @@
...
+}
+
+struct Encoder<'ivm, 'a> {
+  host: &'a Host<'ivm>,
+  table: &'a Table,
+  guide: &'a Guide,
+  grafts: &'a HashMap<NameId, *const Graft<'ivm>>,
+  graft: &'a mut Graft<'ivm>,
+  registers: HashMap<Wire, Register>,
+  active: HashMap<Register, ()>,
```
nilscrm, line 137: I think we can use a `HashSet` here instead. I know we can then not use the entry api but the 8 line match later can then be simplified to
```rs
if !self.active.remove(&reg) {
  self.active.insert(reg);
}
```
which is quite nice and I'm happy to pay one hash lookup for that.

root/derive.vi
```diff
@@ -125,20 +128,21 @@ pub mod Struct {
...
-    Enum::match_(
-      enum_,
-      fn (variant: V) {
-        let result = f(&variant);
-        enum_ = Enum::reconstruct[E, V](variant);
-        result
-      },
-    )
+  pub impl cast_deconstruct[E, V; Enum[E, V]]: Cast[E, V] {
+    fn cast(struct_: E) -> V {
```
nilscrm:
```suggestion
    fn cast(enum_: E) -> V {
```

root/derive.vi
```diff
@@ -125,20 +128,21 @@ pub mod Struct {
...
+      Enum::deconstruct(struct_)
+    }
+  }
+
+  pub impl cast_reconstruct[E, V; Enum[E, V]]: Cast[V, E] {
+    fn cast(content: V) -> E {
+      Enum::reconstruct[E, V](content)
```
nilscrm:
```suggestion
    fn cast(variant: V) -> E {
      Enum::reconstruct[E, V](variant)
```

root/derive.vi
```diff
@@ -100,6 +100,9 @@ pub mod Composite {
   }
 }
 
+pub trait[] NonNil[C] {}
```
nilscrm, line 103: `NonNil` seems too strong of a name for this as it does not includes things like enums, numbers, etc. which are also not nil. Do you think we need this trait more often, otherwise we can inline it into the one place where it is currently being used. Otherwise we can rename the trait to `IsComposite`.

vine/src/components/emitter.rs
```diff
@@ -87,36 +89,38 @@ impl<'a> Emitter<'a> {
...
+      }
+      InterfaceKind::Branch(then, else_) => {
+        let bool = self.emit_port(transfer.data.as_ref().unwrap());
+        self.net.push(TemplateNode::If(*then, *else_, bool, target));
+      }
+      InterfaceKind::Match(enum_id, stages) => {
+        let enum_name = enum_name(self.table, self.guide, self.chart, *enum_id);
+        let enum_ = self.emit_port(transfer.data.as_ref().unwrap());
+        self.net.push(TemplateNode::Match(enum_name, stages.clone(), enum_, target));
+        self.emit_port(transfer.data.as_ref().unwrap());
```
nilscrm: The data is already emitted on line 109.

nilscrm:
I also pushed a commit adding some comments I thought were useful. Have a look at them to see whether my understanding is actually correct.

ivm/src/runtime/ext.rs
enricozb: Would that just increase  (double) the number of split fns we could register? Right now we're limited to `0x3FFF = 16383` of each kind. I think this is likely already enough.

ivy/src/engine.rs
tjjfvi: Ah, I think when I moved the logic to increment `not_nodes` and `not_dirty` to `finish_edit` I forgot to move the `tighten_call`

ivy/src/engine.rs
tjjfvi: This is not currently used but is quite possibly a useful API and I think it's good to have for consistency with the other methods that return an `ExactSizeIterator`

ivy/src/engine.rs
tjjfvi: Actually that shouldn't be done at all, because `from.pending` should be zero

ivy/src/engine/common.rs
tjjfvi: Yep, in fact I originally implemented it that way. However there are cases where a nilary node of type `foo` should erase an n-ary node of type `bar`, but an n-ary node of type `foo` shouldn't commute with an n-ary node of type `bar`. That's why I have `erase`, as it only applies when the lhs is nilary.

ivy/src/engine/common.rs
tjjfvi: This either needs to be a clone, move `aux_` (e.g. `{ aux_ }.find(...)`), or `aux_` needs to be mutable, and I think the `clone()` is cleanest.

ivm/src/runtime/ext.rs
tjjfvi: Yeah, I agree that would be semantically cleaner, but I'm not inclined to change it right now.

ivm/src/host/ext/common.rs
tjjfvi: 🙈

vine/src/components/emitter.rs
tjjfvi: Good catch. I've marked `Wire` as `#[must_use]`, which would've caught this.

ivm/src/host/ext.rs
tjjfvi: I'm essentially simulating if `ExtFn` was a simple tuple struct (like exists for other `Register` things, like `Translate`, `Interaction`, etc.). It just can't be a tuple struct due to some Rust limitations. The function does look a bit weird, but I think it's quite nice to use.

util/src/arithmetic.rs
tjjfvi: This assumes things about the witnesses that doesn't hold in general.

ivm/src/program.rs
tjjfvi: I prefer the current structure because it's consistent with all of the other places where we have a HashMap of wire-like entities and do a `match` on the entry to toggle it (which is a really common pattern in the codebase).

root/derive.vi
tjjfvi: I think it is likely quite generally useful – there are a number of places atm where we have the base case of tuple/object recursion be a singleton, rather than nil, because it's more efficient. Previously, in those cases, I would simply not implement the trait for nil, which isn't great – I think it will be much better to change those to have a `NonNil` bound on the composite implementation, so that the trait can still be implemented for nil.

With regard to naming, IMO it's reasonable – it's a trait implemented for all types which _could be nil but aren't_, so any non-nil tuples and objects.

ivm.iv
nilscrm, line 1: I think this was added by accident

vi.iv
nilscrm, line 1: I think this was added by accident

ivy/src/engine/common.rs
nilscrm: I see, makes sense

root/derive.vi
nilscrm: Fair, agree on the usefulness of the trait. But the name sounds more like the predicate `T != ()` to me which it is not. How about the name `IsComposite`?

tjjfvi: 'composite' == 'tuple or object', so `()` is composite. (The `Composite` trait is somewhat misleading in much the same way `Tuple` and `Object` are, in that it's only implemented for non-nil composites.)

tjjfvi: I would argue that `N32` is not nil, but it is not non-nil, as it is not something for which nil can apply as an adjective, unlike composites, which can be nil or non-nil.

nilscrm (approved):

ivm/src/runtime/ext.rs
```diff
@@ -21,86 +21,66 @@ macro_rules! trait_alias {
...
-      ext_split_fns: Vec::new(),
-      ext_merge_fns: Vec::new(),
-      ext_tys: 1,
-      n32_ext_ty,
-      phantom: PhantomData,
-    }
+impl<'ivm> Extrinsics<'ivm> {
+  /// # Safety
+  ///
+  /// Only one `Extrinsics` may be constructed for a given `'ivm`.
```
enricozb, line 47: Why? B/c of the APIs around getting `ExtFn`'s resulting in IOOB if you mix creation and getting between the two `Extrinsics` structs?

ivm/src/host/ext/common.rs
```diff
@@ -0,0 +1,570 @@
...
+  ) -> impl use<'ivm> + Live<'ivm> + Fn(&mut Runtime<'ivm, '_>, Self) -> ExtVal<'ivm> {
+    let u32 = <u32 as ExtOutput<_>>::register(host, table);
+    move |rt, value| u32(rt, value as u32)
+  }
+}
+
+#[allow(clippy::borrowed_box)]
+// We need to used a borrowed box here since a &[T] would be a fat pointer
+// (ptr + size) which wouldn't fit into a 64-bit Word.
+pub struct Branches<'ivm>(pub &'ivm Box<[*const Graft<'ivm>]>);
```
enricozb, line 398:
```suggestion
// We need to used a borrowed box here since a &[T] would be a fat pointer
// (ptr + size) which wouldn't fit into a 64-bit Word.
#[allow(clippy::borrowed_box)]
pub struct Branches<'ivm>(pub &'ivm Box<[*const Graft<'ivm>]>);
```

ivm/src/host/ext.rs
enricozb: Yeah I'm also a bit uncertain about it. The call sites which register `ExtFn`s look much cleaner for sure. However, I think I would prefer a (proc) macro approach to defining `ExtFn`s, as I think it's a bit hard to follow which path (split/merge & how) a given `ExtFn` goes into. Whereas with a macro you could expand it and see how said function is registered.

I had a branch somewhere that was capable of doing things like this:
```rust
#[ext]
pub fn increment(num: &mut u32) {
  *num += 1
}
```
which would expand to
```rust
pub fn increment<'ivm>() -> crate::host::ext::ExtFnBuilder<'ivm> {
    pub fn increment(num: &mut u32) {
        *num += 1;
    }
    crate::host::ext::ExtFnBuilder {
        name: "increment",
        builder: Box::new(move |host, extrinsics| {
            let transporter_0 = <crate::host::ext::ref_::Ref<
                'ivm,
                u32,
            > as crate::host::ext::ExtTransporterBuilder<
                'ivm,
            >>::register(host, extrinsics);
            let transporter_1 = <() as crate::host::ext::ExtTransporterBuilder<
                'ivm,
            >>::register(host, extrinsics);
            crate::host::ext::ExtFn::Split(
                Box::new(move |ivm, in_0, out_0, out_1| {
                    let Some(mut in_0_0) = crate::host::ext::ExtDecoder::decode(
                        &transporter_0,
                        in_0,
                    ) else {
                        ivm.flags.ext_generic = true;
                        ivm.link_wire(out_0, ivm::port::Port::ERASE);
                        ivm.link_wire(out_1, ivm::port::Port::ERASE);
                        return;
                    };
                    let ret = increment(&mut in_0_0.0);
                    crate::host::ext::ExtWriter::write(
                        &transporter_1,
                        ret,
                        [out_0, out_1],
                        ivm,
                    );
                    in_0_0.done(&transporter_0, ivm)
                }),
            )
        }),
    }
}
```

ivm/src/runtime/flags.rs
```diff
@@ -1,12 +1,13 @@
...
-#[derive(Clone, Copy, Default, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
 pub struct Flags {
   /// A non-copyable extrinsic was copied.
   pub ext_copy: bool,
   /// A non-copyable extrinsic was erased.
   pub ext_erase: bool,
   /// An extrinsic function encountered an unspecified error.
   pub ext_generic: bool,
+  pub invalid_interaction: bool,
```
enricozb, line 10: Should this be printed in the cli, in `RunResult::report`?

ivm/src/runtime/graft.rs
```diff
@@ -129,44 +131,25 @@ impl<'ivm, 'ext> IVM<'ivm, 'ext> {
     }
   }
 
-  /// Execute [`Instructions`], linking the net's root to `port`.
-  pub fn execute(&mut self, instructions: &Instructions<'ivm>, port: Port<'ivm>) {
-    let needed_registers = instructions.next_register.index().max(1);
+  /// Execute a [`Graft`], linking the net's root to `port`.
+  pub fn graft(&mut self, graft: &Graft<'ivm>, port: Port<'ivm>) {
+    let needed_registers = graft.next_register.index().max(1);
```
enricozb: Why is `max(1)` needed here? Isn't a `Graft`'s register field always non-zero?

cli/src/ivm_cli.rs
```diff
@@ -0,0 +1,60 @@
...
+    match <Self as clap::Parser>::parse() {
+      IvmCommand::Run(run) => run.execute(),
+      IvmCommand::Completion(completion) => completion.execute(),
+    }
+  }
+}
+
+#[derive(Debug, Args)]
+pub struct IvmRunCommand {
+  #[arg()]
```
enricozb:
```suggestion
```

ivm/src/system.rs
```diff
@@ -0,0 +1,106 @@
...
+    node.aux.reverse();
+    for _ in goal_inputs..inputs {
+      let next = net.wire();
+      let pair = guide.ext_merge.with_children([guide.pair]);
+      net.add(pair, node.pri, [node.aux.pop().unwrap(), next]);
+      node.pri = next;
+    }
+    for _ in goal_outputs..outputs {
+      let next = net.wire();
+      let ref_ = guide.ext_split.with_children([guide.ref_]);
```
enricozb, line 48: Can you explain a bit why `ref_` is used here? I would expect that outputs would be unpaired with `pair`, but I might be misunderstanding how the outputs are used here.

ivm/src/runtime/ext.rs
tjjfvi: More importantly, because if two different `ExtTyId<'ivm>`s are created for different types, values can get incorrectly transmuted between different types.

ivm/src/runtime/flags.rs
tjjfvi: Good point. I've changed `RunResult::report` to destructure the `Flags`, to ensure exhaustiveness.

ivm/src/runtime/graft.rs
tjjfvi: yeah, I suppose it is unnecessary

ivm/src/system.rs
tjjfvi: If the outputs were unpaired, then it wouldn't be possible to have the outputs be asynchronous, which the API (which gives a wire to each output) allows. By instead using ref to combine the extra output wires with the input, all of the outputs can be asynchronous.

PR #550, opened 2026-03-31
lsp: support wasm compilation target

enricozb:


ivm/src/runtime.rs
```diff
@@ -61,21 +62,34 @@ impl<'ivm, 'ext> Runtime<'ivm, 'ext> {
 
   /// Normalize all nets in this IVM.
   pub fn normalize(&mut self) {
-    let start = Instant::now();
+    self.normalize_hooks(());
+  }
+
+  pub fn normalize_hooks(&mut self, mut hooks: impl Hooks) {
```
tjjfvi: I think I'd prefer `normalize_with` to `normalize_hooks`.

ivm/src/runtime.rs
```diff
@@ -84,9 +98,15 @@ impl<'ivm, 'ext> Runtime<'ivm, 'ext> {
   /// This is useful to get the depth (longest critical path) of the computation
   /// to understand the parallelism of the program.
   pub fn normalize_breadth_first(&mut self) {
-    let start = Instant::now();
+    self.normalize_breadth_first_hooks(());
+  }
+
+  pub fn normalize_breadth_first_hooks(&mut self, mut hooks: impl Hooks) {
```
tjjfvi: similarly here

ivm/src/host/runner.rs
tjjfvi, line 1: this is a host-level api and so shouldn't be in `runtime/`

lsp/src/lib.rs
```diff
@@ -235,12 +231,20 @@ impl Lsp {
...
+    use std::env::current_dir;
+
     let mut path = current_dir().unwrap();
     path.push(&self.file_paths[file]);
     Url::from_file_path(&path).unwrap()
   }
 
+  #[cfg(target_arch = "wasm32")]
+  fn file_to_uri(&self, file: FileId) -> Url {
+    Url::parse(&format!("file://{}", self.file_paths[file].display())).unwrap()
```
tjjfvi, line 244: can this be `Url::from_file_path` instead?

util/src/arithmetic.rs
```diff
@@ -24,153 +24,154 @@ fn Fn<R: Define<I, O, W>, I, O, W>(
...
+
+    pub fn arithmetic<E, $($w,)*>(err: E) -> Arithmetic<E, ($($w,)*)> {
+      Arithmetic(err, PhantomData)
+    }
+  };
+}
+
+arithmetic!(W0, W1, W2, W3, W4, W5, W6, W7, W8, W9, W10, W11, W12, W13, W14, W15, W16, W17, W18);
+
+pub struct Arithmetic<E, WS>(E, PhantomData<fn() -> WS>);
```
tjjfvi: the changes required to this to support the HRTB for the registry are quite sad IMO; here's an alternative that avoids the HRTB: https://github.com/VineLang/vine/commit/acda690de3a21a51a7fc3b20a7bc91e063840fb2

lsp/src/lib.rs
enricozb: `Url::from_file_path` doesn't exist for the wasm32 target.

util/src/arithmetic.rs
enricozb: I wasn't able to get the call-site of `arithmetic` to infer the `R` parameter, but otherwise this works.

tjjfvi (approved):

PR #555, opened 2026-04-21
root: implement full White_Space property for Char::is_whitespace

nilscrm:


tjjfvi:
This doesn't align with the definition Rust uses for ascii whitespace: https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_whitespace

tjjfvi:
Also, perhaps we should rename it to `is_ascii_whitespace`?

nilscrm:
> This doesn't align with the definition Rust uses for ascii whitespace: https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_whitespace

Huh interesting, I was going off of the [`char::is_whitespace`](https://doc.rust-lang.org/src/core/char/methods.rs.html#893) method.

tjjfvi:
sigh, of course `.is_ascii_whitespace() != .is_ascii() && .is_whitespace()`. looks like the only difference is `\x0b` vertical tabulation

nilscrm:
> sigh, of course `.is_ascii_whitespace() != .is_ascii() && .is_whitespace()`. looks like the only difference is `\x0b` vertical tabulation

Don't be ridiculous, that would be way too logical. /s

I've simply extended the method to include the full unicode `White_Space` property, I think that should be the default in any case.

tjjfvi (approved):

PR #556, opened 2026-04-27
vine: fix `#[become]` for generic indirect impls

### LINKED ISSUE #551: vine: fix `#[become]`
It seems that currently `#[become]` doesn't work with certain recursive implementations.

tjjfvi:
Resolves #551 

Previously, an indirect implementation with generic parameters would not be correctly processed for the sake of `#[become]`, which meant that e.g. recursive types with a derived `Fork` or `Drop` would not be replaced with `duplicate`/`erase`.

I've now added a test which explicitly checks that `#[become]` activates when it is supposed to, as well as some tests which have been sped up by `#[become]` functioning properly and replacing `Fork`/`Drop` implementations with `duplicate`/`erase`.


tests/programs/cons.vi
```diff
@@ -0,0 +1,55 @@
+
+use #root::ops::{Cast, Concat, arithmetic::{Mul, Pow}};
+
+#[impl(Show)]
```
nilscrm, line 4: Should an `impl` derivation be allowed if the type also defines a manual implementation of the trait?

nilscrm (approved):

tests/programs/cons.vi
tjjfvi: what do you mean?

nilscrm: Ah nvm for some reason I thought there was an extra implementation of `Show` for `List` but there is not.

nilscrm (approved):

PR #557, opened 2026-04-27
root: zip exact

tjjfvi:
Changes the default zip on iterators from a 'zip shortest' to a 'zip exact', which requires that both iterators have exactly the same length (and otherwise errors). Most of the time when zipping, one expects both lists to have the same length, and them having different lengths is a logic error, so it is nice for that to error rather than be silently ignored. When one does want the zip shortest semantics, they can be used explicitly.

This also allows us to remove the `Drop` bounds from the default zip, allowing it to be used in more circumstances.


tests/programs/iterator_party.vi
```diff
@@ -59,11 +59,18 @@ pub fn main(_: &IO) {
   }
 
   debug::log("--- zip, inspect ---");
+  let a = (0..5).inspect(fn* (&x: &N32) { debug::log("x: {x}") });
+  let b = (0..5).inspect(fn* (&y: &N32) { debug::log("y: {y}") });
+  for t in a.zip_shortest(b) {
```
nilscrm:
```suggestion
  for t in (a, b) {
```

nilscrm (approved):

Issue #533, opened 2026-02-21
vine: type parameter defaults


tjjfvi:
Allow declaring defaults for type parameters. In contexts where inference is not performed (namely, type signatures), omitted type parameters will use the default, if specified. In contexts where inference is performed, the default will be ignored, and the type parameter will be inferred instead.

```rs
trait Add[A, B = A, O = A] { ... }
trait Ord[A, B = A] { ... }

fn foo[T; Add[T], Ord[T]](...) { ... }
// equivalent to
fn foo[T; Add[T, T, T], Ord[T, T]](...) { ... }

struct Array[T, I = N32](...);

fn foo() -> Array[String] { ... }
// equivalent to
fn foo() -> Array[String, N32] { ... }

fn bar() => Array[String, N64] {
  let x = [] as Array[String]; // inferred to be `Array[String, N64]`
  x
}
```

PR #558, opened 2026-04-29
vine: generic argument improvements

tjjfvi:
- When specifying type/impl arguments, it is now valid to omit any number of the arguments, in which case they will be inferred. (Before, it was only allows to omit all or none of the type arguments.) Thus, instead of e.g. `Float::inf[F, _]`, one can now write `Float::inf[F]`.
- Type arguments can now be passed by name (instead of just positionally).

I view this as a precursor to #533.

I considered adding a similar non-positional form to impl arguments (where you could specify what trait the implementation was for), but it added a lot of complexity for little benefit, as impl arguments are very rarely written explicitly.


vine/src/features/generics.rs
```diff
@@ -33,23 +33,38 @@ impl Parser<'_> {
...
     let span = self.start_span();
-    let (name, trait_) = if self.check(Token::Ident) {
-      let path = self.parse_path()?;
-      if path.as_ident().is_some() && self.eat(Token::Colon)? {
-        let trait_ = self.parse_trait()?;
-        (path.as_ident(), trait_)
-      } else {
-        (None, Trait { span: path.span, kind: Box::new(TraitKind::Path(path)) })
-      }
+    let trait_ = self.parse_trait()?;
```
nilscrm, line 36: Since I didn't quite know what this syntax pattern was, I tried looking for names impl parameters in the root library and the tests but I think it's not used anywhere. Might be worth adding a test for.

vine/src/features/generics.rs
```diff
@@ -421,9 +373,73 @@ impl Resolver<'_> {
...
+    &mut self,
+    span: Span,
+    args: &[TypeArg],
+    generics_id: GenericsId,
+    inference: bool,
+  ) -> Vec<Type> {
+    let path = || self.chart.defs[self.chart.generics[generics_id].def].path.clone();
+    let type_param_count = self.sigs.type_params[generics_id].params.len();
+    let mut type_args = vec![None; type_param_count];
+    let mut positional = Counter(0);
```
nilscrm, line 390: I think we should only allow named arguments at the end of the parameter list, otherwise it leads to confusing behavior and error messages.

nilscrm (approved):

PR #559, opened 2026-05-03
vine: rewrite formatter

### LINKED ISSUE #262: vine: fmt: formatter inserting newlines in suboptimal position
Currently the formats the following line:
```rust
pub fn .slice[T, B1, B2; Bound[B1, N32], Bound[B2, N32]](self: List[T], range: Range[B1, B2]) -> List[T] {
```
as follows
```rust
pub fn .slice[T, B1, B2; Bound[B1, N32], Bound[B2, N32]](self: List[T], range: Range[B1, B2]) -> List[
    T,
  ] {
```
which I think is not ideal.

A better option would be for example the following:
```rust
pub fn .slice[T, B1, B2; Bound[B1, N32], Bound[B2, N32]](
  self: List[T],
  range: Range[B1, B2]
) -> List[T] {
```

tjjfvi:
- Resolves #262

The old formatter had a number of limitations – perhaps most notably, it was very 'greedy'; when a syntax node was deciding to be multiline or not, it would only check if it would fit in the max width minus what was written so far onto the line. This means it wouldn't take into account anything *after* it on the same line, and so quite often things would be formatted like
```rs
fn some_function(with: A, really: Long, parameter: List, that: Should, be: Broken) -> List[
  N32,
] {
  // ...
}
```

This occured because when it was printing the parameter list, it only looked at the length of `fn some_function(...)`, which was below the max limit, and so it printed it on a single line. Only once it got to the return type did it realize something was going over the maximum width, so it breaks the return type, which looks really bad.

In the new formatting algorithm, each node receives information about what comes both before and after it on the same line, and decides whether or not to break on that basis. This means that the above example will instead get formatted as
```rs
fn some_function(
  with: A,
  really: Long,
  parameter: List,
  that: Should,
  be: Broken,
) -> List[N32] {
  // ...
}
```
Which is a lot nicer. This happens because the parameter list can see that the return type after it would cause the line to overflow.

If, on the other hand, the parameters were fine, and it was the return type that was super long, it would get formatted like so:
```rs
fn some_function(short: Params, dont: Break) ->
  WoahThisIsAReallyLongReturnType[WhyIsItSoLong]
{
  // ...
}
```

This is because each syntax node can either configure the order in which its children will break, or have it be automatic, in which case larger children will get broken first. Function signatures have the order be automatic, so when the parameter list is excessively long, it's broken, but when the return type is excessively long, it's broken instead.

Other improvements to the formatting include:
- line comments will be respected in all blocks (previously, they would be removed in empty blocks and blocks with only one statement)
- many more syntax elements can be broken (perhaps most notably, binary operators and method chains)
- if expressions can now be single-line
- the maximum width is now respected *precisely* (when possible), as opposed to approximately
  - this includes indent width
- the maximum width can now be configured from the command line

More improvements to the formatter are needed (e.g. respecting comments), but I believe this new formatting algorithm is quite solid, and the remaining changes will mostly be in how the `Content` passed to the algorithm is generated.

I also added a `vine show` command, which prints a Vine file formatted to the width of the terminal and syntax highlighted. (yes I added syntax highlighting to the formatter. no I'm not sorry.) It also has a `--watch` mode which reformats the file as the terminal is resized; it's not _super_ useful from an end-user perspective, but it is fun to play with, and is _very_ useful for debugging/testing the formatter.


tests/aoc_2024/day_03.vi
```diff
@@ -6,10 +6,14 @@ pub fn main(&io: &IO) {
 
   io.println("Part 1: {part1}");
 
-  let part2 = get_muls(input.split("do()").map(fn* (x: String) {
-    let (a, _) = x.split_once("don't()");
-    a
-  }).join(""));
+  let part2 = get_muls(input
+    .split("do()")
```
nilscrm: Instead of having the receiver the previous line and breaking the method call up, let's do the approach we discussed yesterday of only allowing breaking the last element if it is delimited by something.

tests/programs/life.vi
```diff
@@ -145,53 +139,46 @@ pub struct TorusOrthoNeighbors[N, S, W, E]({ n: N, s: S, w: W, e: E });
...
-      s_nw.send(w_se);
-      s_ne.send(e_sw);
-      f(i, j, TorusNeighbors({ nw, n, ne, w, e, sw, s, se }));
-    },
-  )
+  construct_torus_ortho(width, height, fn* (
+    i,
+    j,
+    TorusOrthoNeighbors[N[T], S[T], W[T], E[T]]({ n, s, w, e }),
+  ) {
```
nilscrm: I think this is the only place where I like the old formatting better (not sure it is worth changing)

vine/src/features/if_.rs
```diff
@@ -58,15 +58,18 @@ impl<'src> Formatter<'src> {
     ty: &Option<Ty>,
     then: &Block,
     else_: &Option<Block>,
+    force_multi: bool,
   ) -> Content {
     Content::smart((
```
nilscrm: Should this be `even` like `fmt_expr_if_const`, otherwise we might get the following formatting

```rs
  let x = if                                             
    somethianwdawdawdawdawawdajwhdkahwdjawldkjawljdk     
  { 1 } else { 2 };  
```

tests/programs/array_smoothsort.vi
```diff
@@ -97,11 +97,9 @@ pub fn smooth_sort[T; Ord[T]](&data: &Array[T]) {
...
       let second = second_child(i);
 
-      let (larger_child, child_size) = if data.at(first).assume() < data.at(second).assume() {
-        (second, size - 2)
-      } else {
-        (first, size - 1)
-      };
+      let (larger_child, child_size) = if
+        data.at(first).assume() < data.at(second).assume()
+      { (second, size - 2) } else { (first, size - 1) };
```
nilscrm: Here's an actual example of the other comment

vine/src/features/if_.rs
tjjfvi: Ah, yeah, that's not great. I think `right` is preferable though -- if it suffices to make the branches multiline, we don't want to also force the condition to be multiline.

tests/programs/life.vi
tjjfvi: fixed :)

tests/aoc_2024/day_03.vi
```diff
@@ -6,13 +6,15 @@ pub fn main(&io: &IO) {
...
-    .join("")
+  let part2 = get_muls(
+    input
+      .split("do()")
+      .map(fn* (x: String) {
+        let (a, _) = x.split_once("don't()");
+        a
+      })
+      .join("")
+    ,
```
nilscrm:
```suggestion
```

tjjfvi: This comma was here because this was a multiline argument list, so it had a trailing comma. After discussion, we decided to omit trailing commas in single-argument function calls. (The reasoning being that trailing commas are useful in multiline collections, but a single argument does not function like a collection.)

nilscrm (approved):
very neat

PR #564, opened 2026-05-21
vine: show `fn A(B, C) -> D` instead of `Fn[A, (B, C), D]` in error messages

tjjfvi:
Just a minor QOL change to the diagnostic type printing

### DIFF
--- a/vine/src/structures/types.rs
+++ b/vine/src/structures/types.rs
@@ -390,6 +390,10 @@ impl Types {
     })
   }
 
+  pub(crate) fn is_nil(&self, ty: Type) -> bool {
+    matches!(self.kind(ty), Some((_, TypeKind::Tuple(tys))) if tys.is_empty())
+  }
+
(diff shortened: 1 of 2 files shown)


vine/src/structures/types.rs
```diff
@@ -586,10 +599,33 @@ impl Types {
...
+    *str += "(";
+    let mut first = true;
+    for ty in params {
+      if !first {
+        *str += ", ";
+      }
+      self._show(chart, ty, str);
+      first = false;
+    }
+    *str += ")";
```
enricozb, line 624: could this be simplified if [`Iterator::intersperse`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.intersperse) was stable?

enricozb (approved):

vine/src/structures/types.rs
tjjfvi: possibly yeah

PR #565, opened 2026-05-21
lsp: add lsp tests

enricozb:
Adds a script to mock requests to the LSP and capture a trace of its behavior.


tests/tests.nix
```diff
@@ -166,6 +166,20 @@ let
...
+      result = pkgs.runCommand name { } ''
+        echo "running lsp ${name}"
+        ${pkgs.nushell}/bin/nu ${./lsp.nu} --vine ${vine} --test ${path} >$out
+      '';
+    in
+    {
+      checks."tests-lsp-${name}" = result;
+      snaps."lsp/${name}.log.yml" = result;
+    };
+
```
tjjfvi: personally I'd inline this function since it's not used anywhere else. I only have the `build` and `run` functions extracted up here because they're used multiple times

tjjfvi (approved):

PR #567, opened 2026-06-06
root: json

tjjfvi:
got bored; wrote a json lib

I will eventually want to generalize some of these interfaces (a la serde), but this is nice to have for now.


root/codec/Json.vi
```diff
@@ -0,0 +1,779 @@
...
+              char_ == 'f' { '\x0c' }
+              char_ == 'n' { '\n' }
+              char_ == 'r' { '\r' }
+              char_ == 't' { '\t' }
+              char_ == 'u' {
+                let code = 0;
+                for _ in 0..4 {
+                  assert char is Some(char_) else { return Err("unexpected eof in escape"); }
+                  advance();
+                  code = code * 16;
```
nilscrm:
```suggestion
```

Next line already does `<< 4`

root/codec/Json.vi
```diff
@@ -0,0 +1,779 @@
...
+  fn _parse(str: String) -> Result[Json, String] {
+    let char;
+
+    let fn advance() {
+      char = str!.pop_front();
+    }
+    advance();
+
+    let fn skip() {
+      while char is Some(c) and c <= ' ' and (c == ' ' or c == '\t' or c == '\n' or c == 'r') {
```
nilscrm:
```suggestion
      while char is Some(c) and c <= ' ' and (c == ' ' or c == '\t' or c == '\n' or c == '\r') {
```

root/codec/Json.vi
```diff
@@ -0,0 +1,779 @@
...
+          char_ < ' ' { return Err("unexpected control character in string"); }
+          _ { char_ }
+        })
+      }
+      advance();
+      Ok(str)
+    }
+
+    let fn is_numeric(char: Char) -> Bool {
+      '0' <= char <= '9' or char == '-' or char == '.' or char == '+' or char == 'e'
```
nilscrm:
```suggestion
      '0' <= char <= '9' or char == '-' or char == '.' or char == '+' or char == 'e' or char == 'E'
```

Uppercase E is also allowed

root/codec/Json.vi
```diff
@@ -0,0 +1,779 @@
...
+            assert char is Some(char_) else { return Err("unexpected eof in escape"); }
+            advance();
+            when {
+              char_ == '"' or char_ == '\\' { char_ }
+              char_ == 'b' { '\x08' }
+              char_ == 'f' { '\x0c' }
+              char_ == 'n' { '\n' }
+              char_ == 'r' { '\r' }
+              char_ == 't' { '\t' }
+              char_ == 'u' {
```
nilscrm: This doesn't handle surrogate pairs (see [RFC](https://datatracker.ietf.org/doc/html/rfc8259#section-7)) but that's probably fine for now

root/codec/Json.vi
```diff
@@ -0,0 +1,779 @@
...
+        char == '\r' { "\\r" }
+        char == '\t' { "\\t" }
+        char < '\x10' { "\\u000{char.as[N32].to_hex_raw()}" }
+        _ { "\\u00{char.as[N32].to_hex_raw()}" }
+      }
+    }
+    out ++ "\""
+  }
+
+  pub fn .to_string(json: Json) -> String {
```
nilscrm: Should this maybe be a `Cast[Json, String]` impl

root/codec/Json.vi
```diff
@@ -0,0 +1,779 @@
...
+    }
+  }
+
+  fn _parse(str: String) -> Result[Json, String] {
+    let char;
+
+    let fn advance() {
+      char = str!.pop_front();
+    }
+    advance();
```
nilscrm: I would move this advance to the bottom so that the function definition and the `_parse` body are grouped together

root/codec/Json.vi
```diff
@@ -0,0 +1,779 @@
...
+          advance();
+          Ok(Json::Array(values))
+        }
+        char_ == '"' { Ok(Json::Str(parse_str().try)) }
+        is_numeric(char_) {
+          let num = "";
+          let is_float = false;
+          while char is Some(char_) and is_numeric(char_) {
+            num!.push_back(char_);
+            is_float |= char_ == '.' or char_ == 'e';
```
nilscrm:
```suggestion
            is_float |= char_ == '.' or char_ == 'e' or char_ == 'E';
```

root/codec/Json.vi
```diff
@@ -0,0 +1,779 @@
...
+
+  pub impl bool: ToJson[Bool] {
+    fn to_json(self: Bool) -> Json {
+      Json::Boolean(self)
+    }
+  }
+
+  pub impl list[T; ToJson[T]]: ToJson[List[T]] {
+    fn to_json(self: List[T]) -> Json {
+      Json::Array(self.map(fn* (value: T) { value.to_json() }))
```
nilscrm:
```suggestion
      Json::Array(self.map(ToJson::to_json))
```

root/codec/Json.vi
```diff
@@ -0,0 +1,779 @@
...
+        never as Json
+      }
+    }
+
+    pub impl variant[I, R; ToJsonObject[I], ToJsonVariant[R]]: ToJsonVariant[Variant[I, R]] {
+      fn to_json_variant(variant: Variant[I, R], variant_names: List[String]) -> Json {
+        let (init_name, rest_names) = variant_names.head_tail().assume();
+        match variant {
+          Variant::Init(init) {
+            Json::Object([("type", Json::Str(init_name))] ++ init.to_json_object())
```
nilscrm: I think I prefer the version where enum variants are encoded as
```
{
  "Cycle": { "radius": 1.0 }
}
```
instead of
```
{
  "type": "Circle",
  "radius": 1.0
}
```
Then one can also encode enum variants with non-object content, e.g. `Point(N32, N32)` as
```
{
  "Point": [46, 46]
}
```

root/codec/Json.vi
```diff
@@ -0,0 +1,779 @@
...
+  pub impl nat: FromJson[Nat] = from_integer;
+  pub impl i32: FromJson[I32] = from_integer;
+  pub impl int: FromJson[Int] = from_integer;
+  #[manual]
+  pub impl from_integer[T*; Cast[Int, T], Cast[T, Int]]: FromJson[T] {
+    fn from_json(json: Json) -> Result[T, JsonError] {
+      match json {
+        Json::Integer(int) {
+          let value = int as T;
+          assert int == value as Int else { Err(JsonError::new("integer value out of range")) }
```
nilscrm, line 560: I guess something like `TryCast` would be quite useful here

root/codec/Json.vi
```diff
@@ -0,0 +1,779 @@
...
+  }
+
+  pub impl f32: FromJson[F32] = cast[F64];
+  pub impl f64: FromJson[F64] {
+    fn from_json(json: Json) -> Result[F64, JsonError] {
+      match json {
+        Json::Float(float) { Ok(float) }
+        Json::Integer(int) {
+          let (sign, value) = int.to_parts();
+          let value = if value.parts_len() > 2 { F64::inf } else { value as N64 as F64 };
```
nilscrm: There can be ints with `parts_len() > 2` that are not `inf`, right? If we make `Rational` from `Float::parse` pub we can use `Rational::to_float`.

tjjfvi: true; in my mind since 2^64 > 2^53, it overflowed, but of course that just means it has a higher exponent and can't represent +/- 1 deltas

root/codec/Json.vi
tjjfvi: hmm, I think that encoding is a bit cursed, but it is more general for this case, and it is what serde defaults to...

nilscrm (approved):

PR #570, opened 2026-06-26
lsp: filter diags with none span

enricozb:
- adds `Span::some`
- publishes diagnostics only for non-none spans


tjjfvi (approved):
`Span::NONE` is a bit of a hack, and ideally wouldn't exist, but I suppose it is better for the compiler not to crash if diagnostics are reported against it. (I would though say that most `Span::NONE` diagnostics should be treated as bugs and fixed.)

enricozb:
> I would though say that most Span::NONE diagnostics should be treated as bugs and fixed.)

As bugs in the compiler or bugs in user code? The LSP can't publish these as diagnostics b/c it has no file.

tjjfvi:
Bugs in the compiler, yeah

PR #571, opened 2026-06-29
root: `Lookup` tables

tjjfvi:


vine/src/features/hole.rs
```diff
@@ -40,6 +40,10 @@ impl Finder<'_> {
       let types = types.clone();
       found.push(TypeCtx { types, inner: TirImpl::Synthetic(SyntheticImpl::DefaultNil) });
     }
+    if let Some((Inverted(false), TypeKind::Fn(_))) = types.kind(ty) {
```
enricozb, line 43: How does this work?

I get that `fn` types have sensible defaults (the function they refer to) but I don't understand how the `DefaultNil` impl works here.

Also this could  be combined with the previous if statement.

root/data/Lookup.vi
```diff
@@ -0,0 +1,100 @@
+
+/// Static lookup tables with `O(log(n))` lookup.
+/// ```vi
+/// use #root::data::Lookup;
+/// 
+/// fn primes(_: Lookup::Size8) -> List[N32] {
+///   [2, 3, 5, 7, 11, 13]
+/// }
+///
```
enricozb: ultra-nit: the lines around the `fn primes` item have trailing spaces.

vine/src/features/hole.rs
tjjfvi: values of an fn item type are just nil; the 'function' aspect of them lies in their implementation of the `Fn` trait

tests/programs/lookup.vi
```diff
@@ -6,6 +6,6 @@ pub fn main(&io: &IO) {
   io.println("{squares.lookup(6)}");
 }
 
-fn squares(_: Lookup::Size512) -> List[N32] {
-  (0..512).map(fn* (x: N32) { x * x }).collect[List]()
+fn squares() -> Lookup::Size512[N32] {
+  Lookup((0..512).map(fn* (x: N32) { x * x }).collect[List]())
```
enricozb, line 10: This looks nicer

enricozb (approved):

root/data/Lookup.vi
```diff
@@ -0,0 +1,113 @@
+
+/// Static lookup tables with `O(log(n))` lookup.
+/// ```vi
+/// use #root::data::Lookup;
+///
+/// fn primes(_: Lookup::Size8) -> List[N32] {
+///   [2, 3, 5, 7, 11, 13]
+/// }
```
enricozb, line 8: I think this needs to be updated since the lookup size is in the return param now.

PR #573, opened 2026-07-03
ivm: encode with a stack instead of mutual recursion

enricozb:
Using [`stacker`](https://docs.rs/stacker), allows the vine & ivm cli to run programs which previously caused a stack overflow e.g. (when run with `vine run --opt-all`):

```rust
pub fn main(_: &IO) {
  black_box((0..10000).collect[List]());
}
```

The above program also works with `vine build --opt-all` and `ivm run`.

### DIFF
--- a/ivm/src/program.rs
+++ b/ivm/src/program.rs
@@ -195,6 +195,10 @@ impl<'ivm> Encoder<'ivm, '_> {
   }
 
   fn encode_tree(&mut self, tree: TreeNode, reg: Register) {
+    vine_util::ensure_sufficient_stack(|| self._encode_tree(tree, reg));
+  }
+
+  fn _encode_tree(&mut self, tree: TreeNode, reg: Register) {
     match tree {
       TreeNode::Wire(_) => {}
       TreeNode::Node(name, children) => {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -91,6 +91,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4c95c10ba0b00a02636238b814946408b1322d5ac4760326e6fb8ec956d85775"
 
 [[package]]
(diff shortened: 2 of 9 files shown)


ivm/src/program.rs
```diff
@@ -271,11 +278,19 @@ impl<'ivm> Encoder<'ivm, '_> {
...
 
-  fn encode_binary(&mut self, tag: Tag, label: u16, reg: Register, a: TreeNode, b: TreeNode) {
+  fn encode_binary(
+    &mut self,
+    tag: Tag,
+    label: u16,
+    reg: Register,
+    a: TreeNode,
+    b: TreeNode,
+    stack: &mut Vec<(TreeNode, Register)>,
```
tjjfvi: could this be stored in `&mut Self`?

tjjfvi (commented):
In general it makes me a bit sad to manually rewrite recursive algorithms into something stack based. I might want to consider using something like [`stacker`](https://docs.rs/stacker/latest/stacker/) instead. (Also, I know the Rust compiler uses something like `stacker` to avoid stack overflow in its recursive algorithms; it might be worth looking into what that is and if we could/should use it.)

ivm/src/program.rs
enricozb: Switched to [`stacker`](https://docs.rs/stacker).

enricozb:
> I might want to consider using something like [`stacker`](https://docs.rs/stacker/latest/stacker/) instead.

I updated the PR to have an `ensure_sufficient_stack` in the `vine_util` crate, similar to what [`rust`](https://github.com/rust-lang/rust/blob/c397dae808f70caebab1fc4e11b3edf7e59f58c7/compiler/rustc_data_structures/src/stack.rs#L19) does.

tjjfvi (approved):

